PR #14, opened 2023-03-23
GIVCAMP-33 | Use HeroIcons v2; Port Typography components with clean up

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Use HeroIcon v2
- HeroIcons and Typography components with clean up
- Change folder structure of components: all Storyblok components go into src/components/Storyblok

# Review By (Date)
- Retro

1. Compare preview with code that uses new components
2. Look at code

- JIRA ticket(s) - GIVCAMP-33


src/components/HeroIcon/HeroIcon.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const heroIconStyle = dcnb('su-transition', baseStyle);
+
+  return (
+    <Icon
+      title={title}
+      // If a title for the SVG is provided, unhide the SVG from screen readers
+      aria-hidden={!title}
+      className={dcnb(heroIconStyle, className)}
+      {...props}
+    />
```
yvonnetangsu, line 36: Previously, the Heroicon SVG element is always set to aria-hidden. We made used of the `SrOnlyText` to add any SR text if the icon actually has meaning that needs to be convey to non-sighted users. With HeroIcon v2, we can actually pass a `title` into the SVG which might be a better approach.

PR #28, opened 2023-04-15
GIVCAMP-71 | Add updated Storyblok image service processing with masked CDN URL

yvonnetangsu:
# READY FOR REVIEW

# Summary
- New image utilities that uses the new version of Storyblok Image Service (different from the one we use in SAA Homesite )
- Add WidthBox which we'll need later for x of 12 column type of width options

# Review By (Date)
- Retro

# Review Tasks

1. Look at preview and inspect the card image with the pallas cat
2. See that it uses the new format of the Storyblok image service, and that it uses our masked URL from the assets CDN
Screenshot 2023-04-17 at 9 57 19 AM
Note: image focus is supported and supported image types are converted to webp format for supported browsers
Note: it should preserve the animated GIFS (the pusheen) - it wouldn't get turned into a webp

- JIRA ticket(s) - GIVCAMP-71

### DIFF
--- a/src/utilities/getProcessedImage.ts
+++ b/src/utilities/getProcessedImage.ts
@@ -1,0 +1,47 @@
+import { config } from './config';
+
+/**
+ *
+ * @param {string} image - The Storyblok URL of the image
+ * @param {string} [crop] - The dimension of the image crop (eg., "600 x 400")
+ * @param {string} [focus] - The focal point of the image provided by the Storyblok Asset field (eg, "348x414:349x415")
+ * @param {string} [filters] - Additional filters to apply to the image (eg., "blur(10)").
+ * To add multiple filters, separate them with a colon, eg., "blur(10):grayscale()"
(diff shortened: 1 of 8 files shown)


src/utilities/getProcessedImage.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (filters) {
+    myFilters += `:${filters}`;
+  }
+
+  // The URL of the processed Storyblok image
+  const processedSbUrl = `${imageSrc}${myParams}${myFilters}`;
+  const maskedUrl = processedSbUrl.replace(imageService, `${config.assetCdn}a-us/`);
+
+  return maskedUrl;
+};
```
yvonnetangsu, line 57: This replaces `transformImage` in the SAA repo with updated Storyblok image service and support for the US Storyblok server, also a bunch of logic clean ups

yvonnetangsu:
Pushing this through - please review retro. I might talk about this briefly in adapt dev chat 👀 🤠

PR #33, opened 2023-04-20
GIVCAMP-83 | Update vertical card to include new options

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Add datasource file and add the utilities folder path to TW config so it will generate CSS from any TW classnames in those files.
- Update vertical card component to include the new options

# Review By (Date)
- Retro

# Review Tasks

1. Look at the preview and the code

- JIRA ticket(s) - GIVCAMP-83


tailwind.config.ts
```diff
@@ -22,6 +22,7 @@ export default {
      */
     './src/pages/**/*.{js,jsx,ts,tsx}',
     './src/components/**/*.{js,jsx,ts,tsx}',
+    './src/utilities/**/*.{js,jsx,ts,tsx}',
```
yvonnetangsu, line 25: TW needs to be able to generate classes from datasource which is in utilities folder

yvonnetangsu:
🤠

PR #39, opened 2023-04-25
Nojira | Change base font size

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Base font size for this project. Settled this with @kerri-augenstein today 😄 
- Tried the TW breakpoints but decided to use the Decanter one for now - no change there.

# Review By (Date)
- Retro

# Review Tasks

1. Look at code
2. Check that the new body font size is applied

### DIFF
--- a/src/tailwind/plugins/base/gc-base.js
+++ b/src/tailwind/plugins/base/gc-base.js
@@ -9,6 +9,19 @@ module.exports = function () {
         scrollBehavior: 'smooth',
         overflowY: 'visible !important', // Need this for sticky nav to work
       },
+      body: {
+        fontSize: '1.8rem',
+
+        '@screen md': {
+          fontSize: '1.9rem',
+        },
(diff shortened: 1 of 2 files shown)


src/tailwind/plugins/base/gc-base.js
```diff
@@ -9,6 +9,19 @@ module.exports = function () {
         scrollBehavior: 'smooth',
         overflowY: 'visible !important', // Need this for sticky nav to work
       },
+      body: {
```
sherakama, line 12: Tell me more about this change...

yvonnetangsu: Had double long meetings with Kerri and Ali today. This is the new base font size that Kerri feels like makes the most sense after working on a few modern projects. All the other typography styles are in em unit so will just scale to the base at all breakpoints 😄

PR #40, opened 2023-04-26
GIVCAMP-24 | Release workflow

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Auto release workflow

# Review Tasks

1. Look at code?

- JIRA ticket(s) - GIVCAMP-24

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,103 @@
+# Automatically Create a Release when PR is merged to main with one of the following
+# labels: "patch", "minor", "major", "alpha", "beta", "rc".
+
+name: Auto Release
+
+# Controls when the action will run.
+on:
+  pull_request:
+    types: [ closed ]
+    branches: [ main ]
+
+jobs:
+  release:
+    name: Release
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+
+    # Only run on merged PRs with an appropriate label.
+    if: |
+      github.event.pull_request.merged &&
+      (
+        contains(github.event.pull_request.labels.*.name, 'patch') ||
+        contains(github.event.pull_request.labels.*.name, 'minor') ||
+        contains(github.event.pull_request.labels.*.name, 'major') ||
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,103 @@
...
+        id: release
+        uses: release-drafter/release-drafter@v5
+        with:
+          version: ${{ env.RELEASE_TAG }}
+          publish: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.ADMIN_GITHUB_TOKEN }}
+
+      - name: Merge release back into dev
+        uses: devmasx/merge-branch@v1.4.0
```
yvonnetangsu, line 98: I pretty much stole the one from SAA homesite but updated the version of merge-branch to v1.4.0. Also, removed the reference to the fontawesome license token.

yvonnetangsu:
@sherakama Does this look legit for an autorelease workflow? Copy pasta from SAA repo with a couple changes in the comment above 🤔

sherakama (approved):
This looks good.

If you run in to errors you may have to check/update the github action secret but I think you're good to go there. 

One other thing to note, the release version is based off of the version in package.json so you will be starting at > 0.0.1, which is fine, but something to know.

yvonnetangsu:
> This looks good.
> 
> If you run in to errors you may have to check/update the github action secret but I think you're good to go there.
> 
> One other thing to note, the release version is based off of the version in package.json so you will be starting at > 0.0.1, which is fine, but something to know.

Thanks @sherakama ! What would your recommendation be for the starting version in this case? 😃

sherakama:
0.0.1 is fine by me. 
Some feel strongly it should be 1.0.0-beta|alpha|rc 

I like to save the version 1.0.0 release for go-live. 

Your choice.

yvonnetangsu:
> 0.0.1 is fine by me. Some feel strongly it should be 1.0.0-beta|alpha|rc
> 
> I like to save the version 1.0.0 release for go-live.
> 
> Your choice.

Yup - that sounds like a good plan to me! ⭐

PR #45, opened 2023-04-27
Nojira | Add missing internal link logic to add / in front of cached URL

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Storyblok link fix - for cachedUrl - need to add "/" in front
- Do not pass utm params to external links

# Review By (Date)
- Retro

# Review Tasks

1. Go to the preview and scroll down to the split poster
Screenshot 2023-04-27 at 3 11 40 PM
2. Click on the CTA link button
3. Verify that it scrolls to the page top of the /about-test page
4. On the /about-test page, scroll down to the bottom
5. Click on the red button and verify that it takes you back to home, not /about-test/home

- JIRA ticket(s) - GIVCAMP-
- Other PRs
- Any other contextual information that might be helpful (e.g., description of a bug that this PR fixes, new functionality that it adds, etc.)
- Anyone who should be notified? (`@mention` them here)

### DIFF
--- a/gatsby-node.ts
+++ b/gatsby-node.ts
@@ -11,6 +11,5 @@ exports.createPages = async ({ actions }) => {
   actions.createSlice({
     id: 'global-footer',
     component: path.resolve('./src/components/GlobalFooter/GlobalFooter.tsx'),
-    context: {},
   });
 };
--- a/src/components/Cta/CtaLink.tsx
+++ b/src/components/Cta/CtaLink.tsx
@@ -37,7 +37,7 @@ export const CtaLink = React.forwardRef<HTMLAnchorElement, CtaLinkProps>(
     } = sbLink || {};
 
     // Check for internal links
-    const isInternal: boolean = /^\/(?!\/)/.test(href) || linktype === 'story';
+    const isInternal: boolean = linktype === 'story' || /^\/(?!\/)/.test(href);
 
     // Open internal links in new tab because passing target="_blank" to GatsbyLink doesn't work at the moment
     const openGatsbyLinkInNewTab = (e) => {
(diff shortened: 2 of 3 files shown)


gatsby-node.ts
```diff
@@ -11,6 +11,5 @@ exports.createPages = async ({ actions }) => {
   actions.createSlice({
     id: 'global-footer',
     component: path.resolve('./src/components/GlobalFooter/GlobalFooter.tsx'),
-    context: {},
```
yvonnetangsu, line 14: This doesn't actually do anything so removing

src/components/Cta/CtaLink.tsx
```diff
@@ -79,7 +83,7 @@ export const CtaLink = React.forwardRef<HTMLAnchorElement, CtaLinkProps>(
         {...rest}
         {...sbLinkProps}
         ref={ref}
-        href={myLinkWithUtm}
+        href={myLink}
```
yvonnetangsu, line 86: I originally wanted to pass the utm stuff to external link in case they link out to Giving or Make a Gift, but on second thought, I'll hold off

src/components/Cta/CtaLink.tsx
```diff
@@ -52,6 +52,10 @@ export const CtaLink = React.forwardRef<HTMLAnchorElement, CtaLinkProps>(
     if (isInternal) {
       myLink = cachedUrl || href;
 
+      if (!myLink.startsWith('/')) {
+        myLink = myLink === 'home' ? '/' : `/${myLink}`;
+      }
```
yvonnetangsu, line 57: This was missing which screws with scrolling to top on route change and previously these links renders as relative paths, eg, if I'm on /about and add a Storyblok link with cachedUrl to "test", it will end up linking to /about/test, which isn't what we want. We want it to link to "/test".

src/components/Cta/CtaLink.tsx
sherakama: Yeah, I don't think we want to do that. Usually the codes would be for our sites and specific campaigns but it probably doesn't hurt to add them.

yvonnetangsu: Sounds good to me 😄

yvonnetangsu:
I was scratching my head for a while why it wasn't scrolling to the top on route change, turns out I'm missing the part where I need to add a "/" in front of the cachedUrl if I'm using the Storyblok Link component.

Pushing this through as it works as expected now on the preview 🤠

PR #67, opened 2023-06-02
GIVCAMP-122 | Update play pause buttons in Hero; change intro bracket to outline

yvonnetangsu:
# READY FOR REVIEW

# Summary
- This PR originally was meant to update the animation behavior of the hero. I only include the play/pause button updates because I think the on scroll change isn't good for usability. I'll put that work into a different PR so we can merge this in.
- Use the outline version of the bracket in intro, since the solid one combined with the animation creates visual artifacts that can't seem to go away

# Review Tasks

1. Look at the hero play button inside the bracket as you scroll down, and the background video play pause at the left right corner of the hero

- JIRA ticket(s) - GIVCAMP-122

### DIFF
--- a/src/components/Bracket/Bracket.styles.ts
+++ b/src/components/Bracket/Bracket.styles.ts
@@ -28,7 +28,7 @@ export const curve = (corner: CornerType, isSolid: boolean) => dcnb(
 export const rectangle = (corner: CornerType, isSolid: boolean) => dcnb(
   'su-h-full su-border-y-2',
   {
-    'su-w-[19%] su-bg-current': isSolid,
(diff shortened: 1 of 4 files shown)


src/components/Bracket/Bracket.styles.ts
```diff
@@ -1,7 +1,7 @@
 import { dcnb } from 'cnbuilder';
 
 // Bracket Curve styles
-export const root = 'su-border-current';
+export const root = 'su-border-current su-backface-hidden';
```
yvonnetangsu, line 4: Extra measure to minimize visual artifacts in the brackets due to transform

PR #77, opened 2023-06-09
GIVCAMP-131 | Replace dcnb with cnb for speed

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Replace dcnb with cnb because we don't need the dedupe the way we're using it
- Author says dedupe version is much slower. Looking at the benchmark, we don't get much advantage using it over classname package if we use the dcnb version
https://github.com/xobotyi/cnbuilder/tree/master/benchmark

# Review By (Date)
- Retro

# Review Tasks

1. Look at the preview and see that things still look the same

- JIRA ticket(s) - GIVCAMP-131

### DIFF
--- a/src/components/Bracket/Bracket.styles.ts
+++ b/src/components/Bracket/Bracket.styles.ts
@@ -12,7 +12,7 @@ export const colors = {
 export type ColorType = keyof typeof colors;
 
 export const curve = (corner: CornerType, isSolid: boolean) => cnb(
-  'su-w-[83%] su-h-full su-border-t-2 su-border-l-2 su-rounded-tl-full',
+  'su-w-[83%] su-h-full',
   {
     'su-bg-current': isSolid,
     'su-rounded-tl-full': corner === 'tl',
--- a/src/components/Bracket/Bracket.styles.ts
+++ b/src/components/Bracket/Bracket.styles.ts
@@ -1,4 +1,4 @@
-import { dcnb } from 'cnbuilder';
+import { cnb } from 'cnbuilder';
 
 // Bracket Curve styles
 export const root = 'su-border-current su-backface-hidden';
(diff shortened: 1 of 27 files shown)


src/components/Bracket/Bracket.styles.ts
```diff
@@ -11,8 +11,8 @@ export const colors = {
 };
 export type ColorType = keyof typeof colors;
 
-export const curve = (corner: CornerType, isSolid: boolean) => dcnb(
-  'su-w-[83%] su-h-full su-border-t-2 su-border-l-2 su-rounded-tl-full',
+export const curve = (corner: CornerType, isSolid: boolean) => cnb(
+  'su-w-[83%] su-h-full',
```
yvonnetangsu, line 15: Remove the dupe classes (it was unintentional) here since we aren't using dedupe anymore

sherakama:
Lol. Gave us 2 minutes to review before pulling the trigger?

yvonnetangsu:
> Lol. Gave us 2 minutes to review before pulling the trigger?

Hahaha I didn't think you all would actually have time to review 😂 ! I'll leave them up for longer next time 👍🏼 🤣

sherakama:
Did you measure the difference between `cnb` and `dcnb`? It is supposed to be faster but I am wondering if you ran in to something to trigger this optimization here?

yvonnetangsu:
> Did you measure the difference between `cnb` and `dcnb`? It is supposed to be faster but I am wondering if you ran in to something to trigger this optimization here?

Good question, @sherakama ! 😄  So nope, nothing really trigger this. Just that I don't find us using dedupe at all due to the way we use cnbuilder, so thought why don't I just go with the faster one.

I didn't do benchmarking myself, but comparing the author's results (screenshot below). the cnb version definitely seems more performant. I doubt it's noticeable for us but hey why not 🤷🏼‍♀️ ? With how much animation we're putting in, every little bit helps I guess?

cnb:

Screenshot 2023-06-09 at 12 06 41 PM


dcnb:

Screenshot 2023-06-09 at 12 07 19 PM

sherakama:
Sounds logical to me. Thanks for sharin'

yvonnetangsu:
I ran into a Reddit comment and someone mentioned `classname` vs `clsx` (which is similar to `cnbuilder` but still slower), which made me want to take a look at this again 😄 @sherakama

PR #80, opened 2023-06-14
GIVCAMP-129 | Update color bar in section to extend to superhead

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Update color bar in section to extend to superhead

# Review By (Date)
- Retro

# Review Tasks

1. Look at preview and the /about-test page
2. Check that sections with color bar now extends upward to include the superhead (if there is a superhead)

Screenshot 2023-06-13 at 11 25 25 PM
Screenshot 2023-06-13 at 11 25 15 PM

- JIRA ticket(s) - GIVCAMP-129

### DIFF
--- a/src/components/Storyblok/SbSection.tsx
+++ b/src/components/Storyblok/SbSection.tsx
@@ -47,13 +47,8 @@ export const SbSection = ({
     pb={paddingBottom}
     className="su-relative su-overflow-hidden"
   >
-    {superhead && (
-      <Text size={2} leading="tight" font="serif" className="su-cc su-mb-08em">
-        {superhead}
-      </Text>
-    )}
-    {heading && (
-      <FlexBox className="su-relative su-rs-mb-6">
+    {(heading || superhead) && (
+      <FlexBox className="su-rs-mb-6">
         {barColorValue && (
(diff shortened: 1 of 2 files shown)


src/components/Storyblok/SbSection.tsx
```diff
@@ -47,35 +49,39 @@ export const SbSection = ({
...
+          {heading && (
+            <Heading
+              as={headingLevel}
+              size="splash"
+              leading="none"
+              uppercase
+              font="druk"
+              className="su-mb-0"
+            >
+              {superhead && <SrOnlyText>{`${superhead}:`}</SrOnlyText>}{heading}
```
yvonnetangsu, line 81: Previously recommended by SODA, if we have a superhead above a heading, aria-hidden the visible superhead, and add the superhead as SR only text to the front of the heading.

yvonnetangsu:
It is getting merged 😄

src/components/Storyblok/SbSection.tsx
sherakama: Good to know. This would be a good little bite of info to bring to a dev chat

PR #84, opened 2023-06-20
GIVCAMP-117 | Masthead that becomes visible when scrolling up

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Masthead POC (clients want to see this behavior as recommended by Kerri, kind of a meet in the middle solution for a sticky masthead)
- At the top of the page, the background color of the masthead is transparent, logo and buttons are white
- As one scrolls down, the masthead goes away
- When one scrolls back up, the masthead reappear and stay sticky at the top. However, this version has a solid white background with black logo and buttons.
- Some misc tweaks here and there

# Review By (Date)
- Retro

# Review Tasks

1. Go to the preview
2. Look at the masthead at the top (without scrolling)
3. Scroll down and see that the masthead goes away
4. Scroll further down, then scroll back up to see the white solid version of the masthead reappears as sticky at the top

- JIRA ticket(s) - GIVCAMP-117

### DIFF
--- a/src/components/Homepage/ThemeSection.tsx
+++ b/src/components/Homepage/ThemeSection.tsx
@@ -1,7 +1,7 @@
 import React, { useRef, useState } from 'react';
 import { cnb } from 'cnbuilder';
 import {
-  m, useScroll, useSpring, useTransform, SpringOptions,
+  m, useScroll, useSpring, useTransform, SpringOptions, useWillChange,
 } from 'framer-motion';
 import { Container } from '../Container';
 import { Grid, GridAlternating } from '../Grid';
@@ -34,6 +34,7 @@ export const ThemeSection = ({
     damping: 30,
     restDelta: 0.001,
   };
+  const willChange = useWillChange();
 
   const { scrollYProgress } = useScroll({
     target: containerRef,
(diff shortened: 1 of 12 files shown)


src/components/Homepage/ThemeSection.tsx
```diff
@@ -147,13 +148,13 @@ export const ThemeSection = ({
           <div className="su-max-w-full su-overflow-hidden">
             <m.div
               className="su-h-2 su-bg-white su-origin-left su-mt-300"
-              style={{ scaleX: rightLineSpring }}
+              style={{ scaleX: rightLineSpring, willChange }}
```
yvonnetangsu, line 151: Adding willChange seems to improve flickering on mobile

src/tailwind/plugins/base/gc-base.js
```diff
@@ -6,7 +6,6 @@ module.exports = function () {
   return function ({ addBase, config }) {
     addBase({
       html: {
-        scrollBehavior: 'smooth',
```
yvonnetangsu, line 9: I'm removing smooth scrolling for now. On Gatsby's route change, if someone clicks on a link that's lower on a page, the route change would trigger a scrolling of the new page from that same lower point on the page to the top, which means it would trigger all the animations that depends on scroll position. Also, it's a lot of scrolling movement so probably not great for people who prefers reduced motion either. The only time I can see this being useful is if they click an anchor link on the same page. If we need that later, we'll see if we can add that just for that purpose.

rebeccahongsf: This makes sense to me! We encountered a similar bug that was resolved by removing `scrollBehavior: 'smooth'` on the Alumni Homesite 😄

yvonnetangsu: @rebeccahongsf Good to know. Yeah, I see that as a nice to have for same page anchor links, but I actually don't particularly like the scrolling on every route change. Too much movement 😂

yvonnetangsu:
Merging in as Kerri seems happy for this initial POC. Will finetune after client review and sign off.

PR #89, opened 2023-06-22
GIVCAMP-150 | Update TypeScript to v5 and other related packages

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Update Typescript to latest v5.1.3 - this is a major version upgrade from v4
- Update eslint and other types/typescript related plugins

# Review By (Date)
- When does this need to be reviewed by?

# Review Tasks

1. Look at the preview and confirm that it behaves the same as before
https://deploy-preview-89--giving-campaign.netlify.app/
2. I've tested this locally on Storyblok and it works as before. If you want to double check, you could pull this down and test it out yourself 😄 
3. I've tested running `npm run tsc` and `npm run lint` locally (and in the GitHub action as well) - no problems as well so far. Only thing to keep an eye out for is that `typescript-eslint` officially supports up to Typescript version 5.1.0 and we're using 5.1.3, so it's showing this warning when you run eslint - though I don't think there's any issue for now
Screenshot 2023-06-22 at 4 31 33 PM

- JIRA ticket(s) - GIVCAMP-150


package.json
```diff
@@ -52,14 +52,14 @@
...
+    "@types/react-dom": "^18.2.6",
+    "@typescript-eslint/eslint-plugin": "^5.60.0",
+    "@typescript-eslint/parser": "^5.60.0",
     "autoprefixer": "^10.4.14",
     "decanter": "^7.0.0-rc.4",
-    "eslint": "^8.41.0",
+    "eslint": "^8.43.0",
     "eslint-config-airbnb": "^19.0.4",
     "eslint-config-airbnb-typescript": "^17.0.0",
     "eslint-plugin-import": "^2.27.5",
```
yvonnetangsu, line 65: No updates for these - already the latest

sherakama:
This seems all fine to me. 

You may consider also upgrading to node 20 as that is the current stable.

https://endoflife.date/nodejs

yvonnetangsu:
> This seems all fine to me.
> 
> You may consider also upgrading to node 20 as that is the current stable.
> 
> https://endoflife.date/nodejs

Yeah the TS upgrade was surprisingly smooth. I didn't even need to change anything. 

I'll test out the node upgrade 👍🏼

yvonnetangsu:
> This seems all fine to me.
> 
> You may consider also upgrading to node 20 as that is the current stable.
> 
> https://endoflife.date/nodejs

 @sherakama I might wait a bit since node 20 is still so new. Looks like active LTS for v20 starts Oct this year 👀
https://github.com/nodejs/release

sherakama:
There is no rush, but since you were updating everything thought I'd suggest looking at it.

yvonnetangsu:
> 

I might give it a try and see if it breaks everything 🤣

PR #112, opened 2023-08-19
Nojira | Update nextjs to v13.4.20-canary.12; enable static export; remove use of next image

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Enable static export
- Note: Previously, after updating to anything about next v13.4.9, I couldn't manually go to a page by typing in the URL (using internal links on the site works just fine). This seems to have fixed that problem. Also, previously, going to a non-existent page would show the "error" component, now, it shows the "not-found" component which is preferable.
- Note: TBH I think the app directory still isn't 100% ready for production, but it's good to test things out 🤔 

# Review By (Date)
- Whenever

# Review Tasks

1. Go to the preview and see that all the interaction still works, and it looks the same as before
https://deploy-preview-112--giving-campaign.netlify.app/
2. Manually type `initiatives` after the above URL and hit enter, and check that you're taking to the initiative page instead of an error page
3. Go to https://deploy-preview-112--giving-campaign.netlify.app/monkey and see that it shows the "Not found" temp page.
4. See comments in the code


app/[...slug]/page.tsx
```diff
@@ -14,7 +14,7 @@ type ParamsType = {
   slug: string[];
 };
 
-async function generateStaticParams() {
+export async function generateStaticParams() {
```
Comment (yvonnetangsu) [inline app/[...slug]/page.tsx:17]: To use static export, I need to add `export` back here. Previously it generates a type error, but it seems ok now.

next.config.js
```diff
@@ -1,5 +1,7 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  trailingSlash: true,
+  output: 'export',
```
yvonnetangsu, line 4: To enable static export

netlify.toml
```diff
@@ -2,7 +2,12 @@
 # ##############################################################################
 [build]
   command = "npm run build"
-  publish = ".next"
+  publish = "out"
+
+[build.environment]
+  # For apps that use next export to generate static HTML
+  # set the NETLIFY_NEXT_PLUGIN_SKIP to true.
+  NETLIFY_NEXT_PLUGIN_SKIP="true"
```
yvonnetangsu, line 10: These need to be set to use static export

sherakama: What? Really? Link to Doxn?

yvonnetangsu: Here you go @sherakama
https://nextjs.org/docs/app/building-your-application/deploying/static-exports

It would also show me an error in on the Netlify build page to do these two things

yvonnetangsu: TBH I'm not 100% sure I want to go with this - but at least I finally get it to work once 😂 All the combination of things that blocked this from working before.....

yvonnetangsu:
@sherakama Another thing of interest - comparing deploy time before and after switching to static export + the nextjs update 🤔  ~1.5 min vs ~30s

Screenshot 2023-08-30 at 2 39 31 PM
Screenshot 2023-08-30 at 2 38 54 PM

yvonnetangsu:
@sherakama I merged the latest changes (some not yet in `dev`) into this branch because I created a new `home` page using the new `SbHomepageMvp` component. Merging this change in prevents any future builds in this PR from breaking and allows you to view the new homepage on your local 😄 

Also, I uninstalled the netlify nextjs plugin as you suggested. The build log shows that it now automatically install the latest version - thanks!

yvonnetangsu:
I think I can close this one for now

yvonnetangsu:
@sherakama I previously had this PR which enables static export, and upgraded to 13.4.20-canary.12 which seems to be fine. No Safari button issue and 404 page seems ok too. Tagging you to see if there's anything we can use here.

PR #127, opened 2023-10-03
GIVCAMP-178: Package and node version updates.

sherakama:
# READY FOR REVIEW

# Summary
- Package and node updates
- Lock to 13.4.9 until Netlify can fix their plugin
- Sitemap.xml creation
- Robots.txt creation
- 404 Handling
- Static Rendering
- Consolidated the home page into the catch-all slug file. 
- Upgrade to node 20

# Review By (Date)
- When you can

# Criticality
- Normal

# Review Tasks

1. Review the code
2. Clone down to local
4. Run `nvm use` && `npm run vault:local` && `npm install` && `npm run dev`
5. Make sure the site builds and runs but loading up the home page and then following some links
6. Close down dev environment and run `npm run build` to ensure build works
7. Hop over to the preview build and test that out by loading up the home page and navigating around a few links
8. Navigate to a 404 page eg: https://deploy-preview-127--giving-campaign.netlify.app/this/page/does/not/exist
9. Navigate to the sitemap https://deploy-preview-127--giving-campaign.netlify.app/sitemap.xml
10. Navigate to robots.xml https://deploy-preview-127--giving-campaign.netlify.app/robots.txt


yvonnetangsu:
Got some function crashes when clicking on links (from posters and story cards for example) 😄 I'm sure you noticed that too
Screenshot 2023-10-03 at 12 26 49 PM

.eslintrc.json
```diff
@@ -42,6 +42,7 @@
     "import/no-unresolved": 0,
     "import/export": 0,
     "func-names": 0,
-    "semi": [1, "always"] // 1 is for warning
+    "semi": [1, "always"], // 1 is for warning
+    "@next/next/no-img-element": "off"
```
sherakama, line 46: This was noisy. We don't really need the next/image for optimization but the blur and lazy load is nice. Could we move to using next/image everywhere?

.nvmrc
```diff
@@ -1 +1 @@
-18
\ No newline at end of file
+20
```
sherakama, line 1: It goes to 20!

app/(storyblok)/[...slug]/page.tsx
Comment (sherakama) [inline app/(storyblok)/[...slug]/page.tsx:1]: Moved the storyblok slug pages into a route group so we can add our own layout, error, loading, etc files to just SB pages if we want to add our own static, or dynamic, routes later.

app/(storyblok)/[...slug]/page.tsx
```diff
@@ -1,40 +1,22 @@
-import { Metadata } from 'next';
-import { getStoryblokApi, ISbStoriesParams, StoryblokClient } from '@storyblok/react/rsc';
-import StoryblokStory from '@storyblok/react/story';
+import { Metadata, ResolvingMetadata } from 'next';
+import {
+ getStoryblokApi, ISbStoriesParams, StoryblokClient, StoryblokStory,
+} from '@storyblok/react/rsc';
 import { resolveRelations } from '@/utilities/resolveRelations';
 import { getPageMetadata } from '@/utilities/getPageMetadata';
+import { notFound } from 'next/navigation';
```
Comment (sherakama) [inline app/(storyblok)/[...slug]/page.tsx:7]: This little function triggers the 404 (not-found.tsx) page instead of throwing an error which hits an error boundary and triggers error.tsx.

app/(storyblok)/[...slug]/page.tsx
```diff
@@ -1,40 +1,22 @@
...
 
 const activeEnv = process.env.NODE_ENV || 'development';
 
 type PathsType = {
   slug: string[];
+  fallback: boolean;
```
Comment (sherakama) [inline app/(storyblok)/[...slug]/page.tsx:13]: `{ fallback: false }` means other routes should 404

app/(storyblok)/[...slug]/page.tsx
```diff
@@ -45,29 +27,61 @@ async function getStoryData(params: { slug: string[] }) {
...
-
   const storyblokApi: StoryblokClient = getStoryblokApi();
-  const story = await storyblokApi?.get(`cdn/stories/${slug}`, sbParams);
 
+  try {
+  const story = await storyblokApi?.get(`cdn/stories/${slug}`, sbParams);
   return story;
+  } catch (error) {
+    notFound();
+  }
```
Comment (sherakama) [inline app/(storyblok)/[...slug]/page.tsx:37]: the storyblock client throws an error when it can't find content. Catch it and 404 otherwise you'll get the error.tsx file.

.eslintrc.json
yvonnetangsu: Must be on the same wavelength because I just did that today too 😆 
https://github.com/SU-SWS/ood-giving-campaign/pull/128/commits/93f028a56b4cf1a54c6e74da8357a37171439d92

I tried using the next/image but I don't like how that was a delay, and it doubles up with the Storyblok image services. I'm using the browser native lazy load and that seems to be sufficient (esp now we're using webp and even the large images are small in file size).

yvonnetangsu: Check out these small image file sizes 😄 

Screenshot 2023-10-04 at 12 25 02 AM

Screenshot 2023-10-04 at 12 24 40 AM

app/(storyblok)/[...slug]/page.tsx
Comment (sherakama) [reply app/(storyblok)/[...slug]/page.tsx]: This was only for the `pages` convention. `app` doesn't do this the same way.

sherakama:
@yvonnetangsu 

> Got some function crashes when clicking on links (from posters and story cards for example) 😄 I'm sure you noticed that too Screenshot 2023-10-03 at 12 26 49 PM

Yeah, this is now only crashing on the 404 page, or any non-statically generated page. There is something up with the server-side generation of pages using the app directory, react-server-components, and netlify's runtime. Not sure where or when the issue starts so I'm going to have to spend some time trying to isolate the issue. The alternative is to do what you did in the other PR and go fully static and avoid the netlify/next-runtime altogether. I tried to monkey patch the issue as directed on one of the threads I was reading but that seems to only help a little on nextjs 13.4 and not at all on nextjs 13.5. Some are directing to roll back to nextjs 13.3 😱. I'm planning to start a little debug package and start with just the core next package and build up from there. 

Questions:
- Not sure if the Storyblok integration instructions are really compatible with RSC and Serverless
- Not sure if Framer Motion is causing issues because everything has to be a client component 

Related: 
* https://answers.netlify.com/t/runtime-importmoduleerror-error-cannot-find-module-styled-jsx-style/102375
* https://answers.netlify.com/t/server-edge-not-defined-error-on-nextjs-ssr-functions-cause-site-to-return-500-errors/91793

yvonnetangsu:
> @yvonnetangsu
> 
> Yeah, this is now only crashing on the 404 page, or any non-statically generated page. There is something up with the server-side generation of pages using the app directory, react-server-components, and netlify's runtime. Not sure where or when the issue starts so I'm going to have to spend some time trying to isolate the issue. The alternative is to do what you did in the other PR and go fully static and avoid the netlify/next-runtime altogether. I tried to monkey patch the issue as directed on one of the threads I was reading but that seems to only help a little on nextjs 13.4 and not at all on nextjs 13.5. Some are directing to roll back to nextjs 13.3 😱. I'm planning to start a little debug package and start with just the core next package and build up from there.
> 
> Questions:
> 
> * Not sure if the Storyblok integration instructions are really compatible with RSC and Serverless
> * Not sure if Framer Motion is causing issues because everything has to be a client component
> 
> Related:
> 
> * https://answers.netlify.com/t/runtime-importmoduleerror-error-cannot-find-module-styled-jsx-style/102375
> * https://answers.netlify.com/t/server-edge-not-defined-error-on-nextjs-ssr-functions-cause-site-to-return-500-errors/91793

Thank you for investigating @sherakama! You were working late last night 👀

yvonnetangsu:
@sherakama Is this ready for review? 😄

sherakama:
@yvonnetangsu you can start if you want to but I may change things as I dig in to the SSR / RSC on Netlify issue more.

yvonnetangsu:
> @yvonnetangsu you can start if you want to but I may change things as I dig in to the SSR / RSC on Netlify issue more.

Cool. In that case, I'm going to wait. Going to go heads down to build as much as I can before UComm pull me in. I'll come back to this next week. Thank you so much for this work! 😄

app/(storyblok)/[[...slug]]/page.tsx
Comment (sherakama) [inline app/(storyblok)/[[...slug]]/page.tsx:1]: Moved this to a catch-all slug so we can support more than one level deep slugs

app/(storyblok)/[[...slug]]/page.tsx
```diff
@@ -0,0 +1,116 @@
...
+type PathsType = {
+  slug: string[];
+};
+
+type ParamsType = {
+  slug: string[];
+};
+
+// Control what happens when a dynamic segment is visited that was not generated with generateStaticParams.
+export const dynamicParams = false;
```
Comment (sherakama) [inline app/(storyblok)/[[...slug]]/page.tsx:23]: Setting this to `true` will try to render paths that are not statically generated by fetching the data from Storyblok. Let's wait until Netlify fixes the plugin before we go down that route. Leaving this to false will cause any path not statically generated to 404.

app/(storyblok)/[[...slug]]/page.tsx
```diff
@@ -0,0 +1,116 @@
...
+export async function generateStaticParams() {
+  // Client in the App router is tied to the front end initialization.
+  const storyblokApi: StoryblokClient = getStoryblokApi();
+  let sbParams: ISbStoriesParams = {
+    version: activeEnv === 'development' ? 'draft' : 'published',
+    cv: activeEnv === 'development' ? Date.now() : undefined,
+  };
+
+  // Use the `cdn/links` endpoint to get a list of all stories without all the extra data.
+  const response = await storyblokApi.getAll('cdn/links', sbParams);
```
Comment (sherakama) [inline app/(storyblok)/[[...slug]]/page.tsx:37]: Changed this to `getAll` to support multiple 'pages' of the API.

app/(storyblok)/[[...slug]]/page.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const stories = response.filter((link) => link.is_folder === false);
```
Comment (sherakama) [inline app/(storyblok)/[[...slug]]/page.tsx:38]: Only filter out folders now. We will render the home page with this route as well.

app/(storyblok)/[[...slug]]/page.tsx
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * Fetch the path data for the page and render it.
+ */
+export default async function Page({ params }: { params: ParamsType }) {
+  const { data } = await getStoryData(params);
+
+  // Failed to fetch from API because story slug was not found.
+  if (data === 404) {
+    notFound();
+  }
```
Comment (sherakama) [inline app/(storyblok)/[[...slug]]/page.tsx:109]: Fire up a 404 page if the data was not found.

app/(storyblok)/error.tsx
sherakama, line 1: Default error file that could use some love in the future. Hopefully no one ever sees this.

app/(storyblok)/layout.tsx
sherakama, line 1: Currently, all the layout is inherited from the app route `layout.tsx` file but this file is here so if we wanted to do something specifically for storyblok pages and not other ones we could.

app/(storyblok)/loading.tsx
sherakama, line 1: Default loading shouldn't actually ever be shown as we are statically generating everything.

app/(storyblok)/not-found.tsx
sherakama, line 1: A 404 page that will need some work. We could potentially pull this from Storyblok as well.

app/404/page.tsx
sherakama, line 1: This file is needed because of the way Netlify handles files. I just import from the not-found page to keep it simple.

app/layout.tsx
```diff
@@ -4,6 +4,7 @@ import { Source_Sans_3, Source_Serif_4, Source_Code_Pro } from 'next/font/google
 import localFont from 'next/font/local';
 import { storyblokInit, apiPlugin } from '@storyblok/react/rsc';
 import StoryblokProvider from '@/components/StoryblokProvider';
+import LazyMotionProvider from '@/components/LazyMotionProvider';
```
sherakama, line 7: Moved this provider here, up to the top.

app/loading.tsx
sherakama, line 1: As we are statically rendering this file shouldn't ever be used but it will if we get into ISR or SSR.

app/robots.ts
sherakama, line 1: Simple robots file.

app/sitemap.ts
sherakama, line 1: This sitemap file gets and renders all the paths we statically render.

Not sure why `changeFrequency` or `priority` are not showing up on the render as the documentation claims they were added in 13.4.5. They do work in 13.5.

components/LazyMotionProvider.tsx
sherakama, line 1: Created a Lazymotion provider so that if we wanted to be specific about where we added a client component with this provider we could. Right now we are wrapping the whole app.

components/Image/Image.tsx
sherakama, line 1: Twas empty.

next.config.js
```diff
@@ -8,15 +13,8 @@ const nextConfig = {
...
-  // https://github.com/vercel/next.js/pull/53902
-  modularizeImports: {
-    '@heroicons/react/24/solid': {
-      transform: '@heroicons/react/24/solid/{{member}}',
-    },
-    '@heroicons/react/24/outline': {
-      transform: '@heroicons/react/24/outline/{{member}}',
-    },
+  env: {
+    DEPLOY_PRIME_URL: process.env.DEPLOY_PRIME_URL,
```
sherakama, line 17: Brings this variable to the client. I suppose we could do `window.location` in most cases but this seemed like a good idea for the robots.txt file.

package.json
```diff
@@ -35,24 +35,27 @@
...
     "eslint-config-next": "^13.4.13",
-    "postcss": "^8.4.31",
-    "tailwindcss": "^3.3.3",
-    "typescript": "^5.1.6"
+    "netlify-plugin-vault-variables": "^1.0.7",
+    "postcss": "^8.4.27",
+    "typescript": "^5.1"
+  },
+  "engines": {
+    "node": ">=20.0.0"
```
sherakama, line 59: Expectation of node 20

sherakama:
@yvonnetangsu This is ready for your keen eye.

yvonnetangsu:
> @yvonnetangsu This is ready for your keen eye.

Thank you so much @sherakama. I'll look at this later today (Bambi isn't good for concentration 😂 ) or tomorrow. We can talk more during our 1:1 also.

app/(storyblok)/[...slug]/page.tsx
Comment (yvonnetangsu) [reply app/(storyblok)/[...slug]/page.tsx]: Noice

app/(storyblok)/[[...slug]]/page.tsx
Comment (yvonnetangsu) [reply app/(storyblok)/[[...slug]]/page.tsx:38]: Cool - this works well! I like how you consolidate the 2 pages file to support the "home" page also 😄 

This page which is set as root inside a folder still works 👍🏼 
https://deploy-preview-127--giving-campaign.netlify.app/initiatives

components/Image/Image.tsx
yvonnetangsu: That's me thinking I'll get to this sooner 😂

app/robots.ts
yvonnetangsu: Just a heads up currently I have in the `netlify.toml` that I plan to remove before launch - let me know if this is good practice and if I should do differently.

`X-Robots-Tag = "noindex, nofollow"`

next.config.js
```diff
@@ -1,9 +1,7 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   experimental: {
-    logging: {
-      level: 'verbose',
-    },
+    logging: 'verbose',
```
yvonnetangsu, line 4: @sherakama I updated this to string because I got a warning. I'm guessing you had more options before having to downgrade to 13.4.9? 😄 

Screenshot 2023-10-18 at 12 56 16 PM
Screenshot 2023-10-18 at 12 56 29 PM

package.json
```diff
@@ -35,24 +35,27 @@
...
+    "@types/react-dom": "^18.2",
     "autoprefixer": "^10.4.14",
     "decanter": "^7.1.1",
     "eslint": "^8.50.0",
     "eslint-config-next": "^13.4.13",
-    "postcss": "^8.4.31",
-    "tailwindcss": "^3.3.3",
-    "typescript": "^5.1.6"
+    "netlify-plugin-vault-variables": "^1.0.7",
+    "postcss": "^8.4.27",
```
yvonnetangsu, line 55: Any particular reason that postcss went from ^8.4.31 to ^8.4.27? I'm guess this might be from merging in `dev` (I might've updated this to .31 at some point).

yvonnetangsu (approved):
I pulled this down locally and it all works great. Preview also looks same as before. Thank you for doing a deep dive review of the codebase and added these improvements - it all makes sense. I like how you separate routes for storyblok pages in anticipation of future possible non-Storyblok pages.

Approving this - 🚀 thank you very much @sherakama !

yvonnetangsu:
I have a small question about postcss but I'm going to merge this in so I can start using this in my other PR 😄 🚀

yvonnetangsu:
@sherakama I found a weird issue with this PR on mobile. If I click into one of the stories (Chrome or Safari) on my iPhone from the homepage, I can't use the back button to go back to the homepage. It keeps redirects me back to the story I click on 👀  Everything seems ok on desktop though

sherakama:
> @sherakama Clarifying question - we're only going for static rendering, but not static export (SSG) like in #112 ? I'm assuming that's in anticipation of preview/draft mode which probably isn't supported by static export? 😄

I'm keeping my options open for SSR on the preview mode. I suspect we could scaffold something like in the Gatsby builds where we cobble together an editor page through the storyblok bridge to support client side rendering if we went full static. I'm hoping to use the 'latest and greatest' from storyblok and take advantage of RSC where I can.

package.json
sherakama: With the version constraint, those patch versions won't matter as they will still pull the latest patch. Not sure where that difference snunk in.

yvonnetangsu:
> > @sherakama Clarifying question - we're only going for static rendering, but not static export (SSG) like in #112 ? I'm assuming that's in anticipation of preview/draft mode which probably isn't supported by static export? 😄
> 
> I'm keeping my options open for SSR on the preview mode. I suspect we could scaffold something like in the Gatsby builds where we cobble together an editor page through the storyblok bridge to support client side rendering if we went full static. I'm hoping to use the 'latest and greatest' from storyblok and take advantage of RSC where I can.

Makes sense! 👍🏼 😄

PR #128, opened 2023-10-04
GIVCAMP-220 | Grid options and story card section layouts

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Option to right align Section header
- Add options for grid 
- Create layout for Story Cards on the homepage

# Review By (Date)
- Retro

# Review Tasks

1. go to https://deploy-preview-128--giving-campaign.netlify.app/
2. Look at the 1, 2, 3, 4 up story card sections
3. Look at the different breakpoints

- JIRA ticket(s) - GIVCAMP-220

### DIFF
--- a/components/Grid/Grid.styles.ts
+++ b/components/Grid/Grid.styles.ts
@@ -50,6 +50,8 @@ export const gridNumCols = {
   },
 };
 
+export const rtl = '[direction:rtl] children:[direction:ltr]';
+
 export const gridGaps = {
   default: 'grid-gap',
   card: 'grid-gap gap-y-50 xl:gap-y-70',
@@ -92,7 +94,7 @@ export const gridAlignItems = {
 };
 
 // Styles for the alternating grid
-export const alternatingGridWrapper = 'relative mx-auto';
+export const alternatingGridWrapper = 'relative mx-auto cc';
 
 export const centerline = 'hidden md:block w-2 absolute top-0 left-1/2 -ml-1 h-full origin-top bg-current';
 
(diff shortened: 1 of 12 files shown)


yvonnetangsu:
I might merge this one in so client can discuss with us tomorrow. We might tweak this further but I'm going to do that in a future PR, so that I can continue to build out the homepage 😄

components/Grid/Grid.styles.ts
```diff
@@ -50,6 +50,8 @@ export const gridNumCols = {
   },
 };
 
+export const rtl = '[direction:rtl] children:[direction:ltr]';
```
yvonnetangsu, line 53: This trick makes the grid row items fill from right to left, but need to change the direction of the children from left to right. It's too bad we don't have a CSS property that's kind of like flex row-reverse.

components/Storyblok/SbGrid.tsx
```diff
@@ -1,11 +1,16 @@
...
 import { Grid, type GridGapType, type GridNumColsType } from '../Grid';
+import { WidthBox, type WidthType } from '../WidthBox';
 import { CreateBloks } from '../CreateBloks';
 
 type SbGridProps = {
   blok: {
     _uid: string;
     gap?: GridGapType;
+    boundingWidth?: 'site' | 'full';
+    width?: WidthType;
```
yvonnetangsu, line 11: Slight tweak to Grid options to have `boundingWidth` and `Width`

boundingWidth: the width of the bounding box that is the outer container, either edge-to-edge or centered-container

width: within the bounding box, number of columns that the grid takes up. Eg, it can be 6 of 12 columns of a centered container, or 8 of 12 columns of an edge-to-edge container.

PR #134, opened 2023-10-09
Fix | Margin fix when no bar color is selected for Sections

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Fixup for Sections with no bar color options selected

# Review By (Date)
- Retro

# Review Tasks

1. Go to the preview and look at all the Sections now with no color bars selected and check that they're align to the edge of the centered container (left or right side depending on whether the section header is left or right aligned)
Screenshot 2023-10-09 at 11 54 12 AM
Screenshot 2023-10-09 at 11 54 02 AM


yvonnetangsu:
Pushing this one through so client can see the update. It's small fix.

sherakama:
smol

PR #150, opened 2023-10-31
GIVCAMP-244 | Social sharing

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Story social sharing widget

# Review By (Date)
- Retro

# Review Tasks

1. Go to https://deploy-preview-150--giving-campaign.netlify.app/stories/education-for-a-life-of-purpose
3. See the social sharing widget at the top and bottom of the story
Screenshot 2023-10-30 at 9 50 00 PM
Screenshot 2023-10-30 at 9 50 06 PM
3. Check all responsive behavior
4. Click the first button and see that it copies the production URL (domain TBD but I'm using the temporary one https://impactstories.stanford.edu/)
5. Check that other buttons work (the Linkedin one is not working with this fake URL)
3. Go to this story https://deploy-preview-150--giving-campaign.netlify.app/stories/from-the-atom-to-the-universe
6. See how I can disable the widget at the top, so only the one at the bottom of the story is showing (there is no story content here yet)

- JIRA ticket(s) - GIVCAMP-244


components/SocialSharing/SocialSharing.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <Heading as="h2" font="sans" size="base" weight="semibold" leading="none" className="mb-0">
+            Share this story
+          </Heading>
+          <FlexBox className="gap-6 md:gap-12">
+            <SocialButton
+              onClick={handleCopyClick}
+              aria-label={buttonState === 'copied' ? 'URL has been copied' : 'Copy story URL'}
+            >
+              {buttonState === 'copied' ? <HeroIcon icon="copy" /> : <LinkIcon />}
+            </SocialButton>
```
yvonnetangsu, line 49: @sherakama Got a use for the slug that I passed from the Story here 😄

components/SocialSharing/FacebookF.tsx
```diff
@@ -0,0 +1,7 @@
+export const FacebookF = ({ ...props}: React.SVGProps<SVGSVGElement>) => (
+  <svg {...props} xmlns="http://www.w3.org/2000/svg" viewBox="0 0 320 512" className="fill-current">
+    {/*!-- Font Awesome Free 6.4.2 by @fontawesome - https://fontawesome.com License -
+    https://fontawesome.com/license (Commercial License) Copyright 2023 Fonticons, Inc. --*/}
+    <path d="M279.14 288l14.22-92.66h-88.91v-60.13c0-25.35 12.42-50.06 52.24-50.06h40.42V6.26S260.43 0 225.36 0c-73.22 0-121.08 44.38-121.08 124.72v70.62H22.89V288h81.39v224h100.17V288z" />
+  </svg>
+);
```
yvonnetangsu, line 7: Using FontAwesome is really annoying on Nextjs
https://fontawesome.com/docs/web/use-with/react/use-with

I tried installing the React packages, and it didn't work out of the box of course.  Found out I have to do a bunch of ugly workarounds. It's simply not worth it for 5 icons so I copied the SVG on the site and convert them into JSX. Much easier this way and doesn't require weird workarounds.

components/SocialSharing/SocialSharing.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  const handleCopyClick = () => {
+    copy(`${config.siteUrlProd}/${slug}`);
+    setButtonState('copied');
+
+    // Reset the button state after 5 seconds
+    setTimeout(() => {
+      setButtonState('default');
+    }, 5000);
+  };
```
yvonnetangsu, line 40: After 5 seconds, the copy link button resets itself to use the original icon.

components/SocialSharing/SocialSharing.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  const safeUrl = encodeURIComponent(`${config.siteUrlProd}/${slug}`);
```
yvonnetangsu, line 42: Make the story URL safe for sharing

components/SocialSharing/SocialSharing.tsx
```diff
@@ -0,0 +1,90 @@
...
+            </SocialButton>
+            <SocialButton
+              aria-label="Share on X (formerly Twitter)"
+              href={`${twitterBaseURL}${safeUrl}${title ? `&text=${title}` : ''}`}
+            >
+              <TwitterX aria-hidden width="18" />
+            </SocialButton>
+            <SocialButton
+              aria-label="Share on LinkedIn"
+              href={`${linkedinBaseURL}${safeUrl}`}
```
yvonnetangsu, line 81: It's weird on Linkedin currently, but my theory is that it doesn't like the URL because we don't actually have that domain at the moment 😬

utilities/config.ts
```diff
@@ -2,8 +2,8 @@
  * Global variables for this project.
  */
 export const config = {
-  siteUrlProd: 'https://stanfordonpurpose.edu',
-  siteTitle: 'Stanford On Purpose',
+  siteUrlProd: 'https://impactstories.stanford.edu',
+  siteTitle: 'Stanford Impact Stories',
```
yvonnetangsu, line 6: Just guessing at this point. Nothing confirmed yet

package.json
```diff
@@ -37,7 +37,8 @@
     "framer-motion": "^10.15.0",
     "next": "13.4.9",
     "react-player": "^2.13.0",
-    "storyblok-rich-text-react-renderer-ts": "^3.2.0"
+    "storyblok-rich-text-react-renderer-ts": "^3.2.0",
+    "usehooks-ts": "^2.9.1"
```
yvonnetangsu, line 41: Installed this package since it has some useful hooks which I might need later. I'm using the copy to clipboard one for this PR

sherakama: I like this package.

components/SocialSharing/SocialSharing.tsx
sherakama: These are all client components. Wouldn't `window.location` suffice? Nothing wrong with prop drilling the slug I suppose, but I am wondering what the overall benefit is?

components/SocialSharing/SocialSharing.tsx
sherakama: neat

sherakama (approved):

components/SocialSharing/SocialSharing.tsx
yvonnetangsu: Yeah, I think window.location probably should work here. I think I had painful memory associated with getting the metadata to work so my brain naturally avoided that and went with passing the slug 😂

yvonnetangsu:
@sherakama Thank you for the quick review! 🎉 😄 I kinda of made up the hocus states so might add tweaks later if designers want changes.

PR #152, opened 2023-10-31
GIVCAMP-195: Typeform embed options.

sherakama:
# READY FOR REVIEW

# Summary
- Typeform embedding options

# Review By (Date)
- Whenever

# Criticality
- ?

# Review Tasks

1. Review code
2. Navigate to https://deploy-preview-152--giving-campaign.netlify.app/test/typeform
3. Check this branch out on your local and fire up the https proxy
4. [Try out the new block in Storyblok](https://app.storyblok.com/#/me/spaces/1005200/stories/0/0/508520?env=PR+%23152) (Use preview URL # 152) and on your localhost
5. Review Storyblok options, labels, and descriptions
6. Review styles

- JIRA ticket(s) - GIVCAMP-195


app/typeform/pop-over-full/page.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { PopOver } from '@/components/TypeForm/PopOver';
+
+const TypeFormPage = () => {
```
yvonnetangsu, line 4: That was smart earlier that you separated the storyblok pages into a different storyblok route - thanks @sherakama !

yvonnetangsu (approved):
Thank you for this PR, @sherakama! This works great 🚀  I guess we could just merge this in, and remove what we don't need later on.

Only thing I can think of is that I'm using cnb instead of dcnb and thus no need for the className: ClassValue type. With the way we use TW, we never ended up needing dedupe (it never quite work the way I want it to) and the dcnb is slower than cnb so I switched to that when I started the project 😄

sherakama:
> Thank you for this PR, @sherakama! This works great 🚀 I guess we could just merge this in, and remove what we don't need later on.

Let me move the needle on this a bit more first. I'll do a little refactoring to make this into Storyblok components and editable so that the author can add their form id. 
 
> Only thing I can think of is that I'm using cnb instead of dcnb and thus no need for the className: ClassValue type. With the way we use TW, we never ended up needing dedupe (it never quite work the way I want it to) and the dcnb is slower than cnb so I switched to that when I started the project 😄

Sure. Sounds good.

yvonnetangsu:
Perfect - I'll let you merge in yourself when you feel it's ready 👍🏼

sherakama:
@yvonnetangsu This is ready for you again. 

I've added some basic styling and a header. The options in the Storyblok UI cover a lot of the Typeform SDK so there should be plenty to try out.

components/Storyblok/SbTypeform.tsx
```diff
@@ -0,0 +1,162 @@
...
+    displayAsFullScreenModal: boolean;
+  };
+};
+
+// TODO: Move this to the non-storyblok Typeform component dir.
+export const styles = {
+  container: 'cc flex flex-col place-items-center rs-py-6',
+  heading: 'px-[40px] tf-sm:px-[80px] max-w-[878px]',
+  fauxCTA: cnb(CTAStyles.cta, CTAStyles.ctaVariants.solid, CTAStyles.ctaSizes.large),
+};
```
sherakama, line 35: Some basic styles. We can move this somewhere else if/when things change.

components/Storyblok/SbTypeform.tsx
```diff
@@ -0,0 +1,162 @@
...
+            id={id}
+            autoClose={1000}
+            disableScroll={disableScroll}
+            enableSandbox={enableSandbox}
+            hideFooter={hideFooter}
+            hideHeaders={hideHeader}
+            transitiveSearchParams={transitiveParams}
+            displayAsFullScreenModal={displayAsFullScreenModal}
+          >
+            <span className={styles.fauxCTA}>{buttonLabel}</span>
```
sherakama, line 77: I faked the CTA styles for this embed. Feel free to swap it out with something else.

components/TypeForm/PopOver.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const [isClient, setIsClient] = useState(false);
+  useEffect(() => {
+    setIsClient(true);
+  }, []);
+
+  if (!isClient) {
+    return null;
+  }
+
+  return <Popover noHeading keepSession id={id} {...rest} />;
```
sherakama, line 20: No heading text and the session is maintained if the widget gets refreshed (webpack HMR or window reload)

components/TypeForm/SideTab.tsx
```diff
@@ -0,0 +1,26 @@
...
+}:SideTabProps) => {
+
+  const [isClient, setIsClient] = useState(false);
+  useEffect(() => {
+    setIsClient(true);
+  }, []);
+
+  if (!isClient) {
+    return null;
+  }
```
sherakama, line 21: Protection against the app dir. The SDK isn't RSC ready.

tailwind/plugins/theme/gc-screens.js
```diff
@@ -8,6 +8,7 @@ module.exports = function () {
   return {
     '3xl': '1700px',
     '4xl': '2000px',
+    'tf-sm': '658px',
```
sherakama, line 11: The TF breakpoints are a little different than ours.

yvonnetangsu: Interesting..... 😄

yvonnetangsu:
Did some exploration with the settings. While transparent background isn't in the theme panel, it is in the "share" settings where one can adjust the % of background transparency - I forgot to take a screen cap before they very quickly took away those features after the initial preview since I'm on a free plan. You can see it here:
https://community.typeform.com/build-your-typeform-7/transparent-background-3509

components/Storyblok/SbTypeform.tsx
yvonnetangsu: I will likely take away some of the layout styles here later on, since I'm going to expect this form to be nested inside a Section, then a Grid (which provides option for bounding width - cc or edge to edge, and # of columns within the bounding box). Happy to leave these in for now - I'll get to the stylings later 😄

yvonnetangsu:
Actually looks like there's an opacity prop - we can add it later if needed
https://github.com/Typeform/embed/blob/HEAD/packages/embed/README.md#options
Screenshot 2023-11-06 at 11 35 07 AM

yvonnetangsu:
Another concern is accessibility of some of the options - it looks like they do try to make it accessible, but some of the UX is slightly different from what I'm normally used to, eg, if I open the drawer (slider option, if I hit tab after it shows the "Sign up" button, it just goes to the email text field without me having to hit enter). Perhaps we keep these options but let OOD knows not to use them before we confirm with SODA that they're accessible @sherakama 😄

yvonnetangsu (approved):
Other than question about whether some of the props can be optional and that we'll continue to finetune on options and styling, these examples seem to work well! Thank you for exploring this package @sherakama! 😄 🚀 

Please feel free to merge when you're ready. Perhaps not advertise some of the option before we confirm with SODA that they're accessible, eg, slider. Thanks!

components/Storyblok/SbTypeform.tsx
```diff
@@ -0,0 +1,162 @@
...
+    fullScreen: boolean;
+    heading: string;
+    hideFooter: boolean;
+    hideHeader: boolean;
+    buttonLabel: string;
+    embedHeight: string;
+    disableScroll: boolean;
+    enableSandbox: boolean;
+    transitiveSearchParams: string;
+    displayAsFullScreenModal: boolean;
```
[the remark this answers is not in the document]
sherakama: Yeah, let me have another go at that.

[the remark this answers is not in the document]
sherakama: ah no, there are 'required' as this is essentially the schema that Storyblok will send us every time. Downstream the components are optional but this is the expected shape of the API response.

sherakama:
> Did some exploration with the settings. While transparent background isn't in the theme panel, it is in the "share" settings where one can adjust the % of background transparency - I forgot to take a screen cap before they very quickly took away those features after the initial preview since I'm on a free plan. You can see it here: [community.typeform.com/build-your-typeform-7/transparent-background-3509](https://community.typeform.com/build-your-typeform-7/transparent-background-3509)

Ah, neat. I've added a 'transparent' option to the Storyblok component as a boolean. When set, it will set the opacity to either 100 or 0.

sherakama:
> Another concern is accessibility of some of the options - it looks like they do try to make it accessible, but some of the UX is slightly different from what I'm normally used to, eg, if I open the drawer (slider option, if I hit tab after it shows the "Sign up" button, it just goes to the email text field without me having to hit enter). Perhaps we keep these options but let OOD knows not to use them before we confirm with SODA that they're accessible @sherakama 😄


Let me test the autofocus option and see if it helps. If it does, I'll force it on that component.

sherakama:
> Let me test the autofocus option and see if it helps. If it does, I'll force it on that component.

Not an option on that widget. Will have to pass that through to SODA.

yvonnetangsu:
> > Let me test the autofocus option and see if it helps. If it does, I'll force it on that component.
> 
> Not an option on that widget. Will have to pass that through to SODA.

No worries - we likely won't need these options for MVP. If they're interested, we can tell them it needs to go through a11y review first and we can wait for post MVP since we need to focus on the other things first 😄

PR #153, opened 2023-11-01
GIVCAMP-237 | Story Hero caption

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Support for caption for hero image(s)

# Review By (Date)
- Retro

# Review Tasks

1. Go to https://deploy-preview-153--giving-campaign.netlify.app/stories/education-for-a-life-of-purpose
2. See the caption appear below the hero, in this case, below the textured bar that is below the hero
Screenshot 2023-11-01 at 10 58 02 AM
3. Go to this story
https://deploy-preview-153--giving-campaign.netlify.app/stories/the-most-wicked-problems
4. See that there's a caption there below the hero too (this time no textured bar)

- JIRA ticket(s) - GIVCAMP-237

### DIFF
--- a/components/BlurryPoster/BlurryPoster.tsx
+++ b/components/BlurryPoster/BlurryPoster.tsx
@@ -43,6 +43,10 @@ type BlurryPosterProps = HTMLAttributes<HTMLDivElement> & {
   alt?: string;
   tabColor?: AccentColorType;
   cta?: React.ReactNode;
+  /**
+   * This isn't displayed in the component, aria-describedby will be added to the images if a caption exists
+   */
+  caption?: string;
 };
 
 export const BlurryPoster = ({
@@ -69,6 +73,7 @@ export const BlurryPoster = ({
   alt,
   tabColor,
   cta,
+  caption,
   className,
   ...props
 }: BlurryPosterProps) => {
@@ -84,11 +89,13 @@ export const BlurryPoster = ({
       <img
         src={getProcessedImage(bgImageSrc, '1200x1200', bgImageFocus)}
         alt={bgImageAlt || ''}
+        aria-describedby={caption ? 'story-hero-caption' : undefined}
         className={styles.bgImageMobile}
       />
       <img
         src={getProcessedImage(bgImageSrc, '2000x1200', bgImageFocus)}
         alt={bgImageAlt || ''}
+        aria-describedby={caption ? 'story-hero-caption' : undefined}
         className={styles.bgImage}
       />
       <div className={styles.blurWrapper(addBgBlur, addDarkOverlay, type, bgColor)}>
@@ -176,11 +183,13 @@ export const BlurryPoster = ({
                 <img
                   src={getProcessedImage(imageSrc, type === 'hero' && !isTwoCol ? '1800x900' : '900x1200', imageFocus)}
                   alt={alt || ''}
+                  aria-describedby={caption ? 'story-hero-caption' : undefined}
                   className={styles.image}
                 />
                 <img
                   src={getProcessedImage(imageSrc, type === 'hero' ? '1000x1000' : '1000x500', imageFocus)}
                   alt={alt || ''}
+                  aria-describedby={caption ? 'story-hero-caption' : undefined}
                   className={styles.imageMobile}
                 />
               </AnimateInView>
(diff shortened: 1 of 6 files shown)


components/BlurryPoster/BlurryPoster.tsx
```diff
@@ -84,11 +89,13 @@ export const BlurryPoster = ({
       <img
         src={getProcessedImage(bgImageSrc, '1200x1200', bgImageFocus)}
         alt={bgImageAlt || ''}
+        aria-describedby={caption ? 'story-hero-caption' : undefined}
```
yvonnetangsu, line 92: This is the background image in the hero (optional). It shares the same caption with the foreground hero image (optional)

components/BlurryPoster/BlurryPoster.tsx
```diff
@@ -176,11 +183,13 @@ export const BlurryPoster = ({
                 <img
                   src={getProcessedImage(imageSrc, type === 'hero' && !isTwoCol ? '1800x900' : '900x1200', imageFocus)}
                   alt={alt || ''}
+                  aria-describedby={caption ? 'story-hero-caption' : undefined}
```
yvonnetangsu, line 186: This is the foreground image

components/Hero/StoryHeroMvp.tsx
```diff
@@ -140,10 +141,24 @@ export const StoryHeroMvp = ({
...
       )}
+      {caption && (
+        <Container bgColor="white" className={styles.captionWrapper}>
+          <Text
+            // id is for aria-describedby in the images since we can't use figcaption here
+            id="story-hero-caption"
+            variant="caption"
+            leading="display"
+            className={styles.caption}
+          >
```
yvonnetangsu, line 157: Add an id to the caption so it can be referenced by `aria-describedby` in the hero foreground and background images. With how the hero is structured, we can't use the figure/figcaption HTML tags.

components/BlurryPoster/BlurryPoster.tsx
```diff
@@ -43,6 +43,10 @@ type BlurryPosterProps = HTMLAttributes<HTMLDivElement> & {
   alt?: string;
   tabColor?: AccentColorType;
   cta?: React.ReactNode;
+  /**
+   * This isn't displayed in the component, aria-describedby will be added to the images if a caption exists
+   */
+  caption?: string;
```
sherakama, line 49: Why not label this `ariadescribedby` then?

components/Hero/StoryHeroMvp.tsx
sherakama: Will there only ever be one of these on a page? If there might be more than one, you will want a dynamic id.

yvonnetangsu: @sherakama Yup, there will always only be one single caption for the hero, so not too concerned about the uniqueness here 😄

components/BlurryPoster/BlurryPoster.tsx
yvonnetangsu: The caption isn't actually the text that is displayed in the BlurryPoster (doubles up as the top part of the story hero sans caption/textured bar) nor for the `aria-describedby` attribute. I'm only passing the caption prop in (if there is a caption passed from the StoryMVP page) so I can do a boolean check on whether a caption exists. If there is one, I add `aria-describedby="story-hero-caption"` (the id of the caption which is always unique) to all the images in the hero.

yvonnetangsu: Like this - just to tell the screen reader that the hero image has a caption associated with it.

Screenshot 2023-11-01 at 1 56 09 PM

components/Hero/StoryHeroMvp.tsx
sherakama: Will there only ever be one hero on the page?

yvonnetangsu: Yup, only one hero ever 😄 It's not a nestable blok that can be dropped in anywhere.

components/BlurryPoster/BlurryPoster.tsx
sherakama: ah thanks for splainin.

components/Hero/StoryHeroMvp.tsx
sherakama: Ok. Great.

yvonnetangsu:
Merging this in @sherakama - thank you for reviewing and keeping me honest 🤠

PR #158, opened 2023-11-03
GIVCAMP-223 | Story image; layout component options; packages updates

yvonnetangsu:
# READY FOR REVIEW

# Summary
- image component + various updates in other components
- Update Decanter and other packages

# Review By (Date)
- Retro
# Review Tasks

1. https://deploy-preview-158--giving-campaign.netlify.app/stories/charging-ahead-batteries-of-the-future
2. https://deploy-preview-158--giving-campaign.netlify.app/stories/the-second-envelope

Look at the two previews, and see all the image component options, plus layouts such as these 50/50 splits

Screenshot 2023-11-06 at 6 26 25 PM

Check responsive and make sure that the image fills the grid cell when the quote side is longer
Screenshot 2023-11-06 at 6 26 11 PM

Screenshot 2023-11-06 at 6 58 47 PM

- JIRA ticket(s) - GIVCAMP-223

### DIFF
--- a/components/Container/Container.tsx
+++ b/components/Container/Container.tsx
@@ -4,6 +4,10 @@ import {
   paddingTops,
   paddingBottoms,
   paddingVerticals,
+  marginTops,
+  marginBottoms,
+  marginVerticals,
+  type MarginType,
   type PaddingType,
 } from '@/utilities/datasource';
 import * as styles from './Container.styles';
@@ -15,6 +19,9 @@ export type ContainerProps = HTMLAttributes<HTMLElement> & {
   pt?: PaddingType;
   pb?: PaddingType;
   py?: PaddingType;
+  mt?: MarginType;
+  mb?: MarginType;
+  my?: MarginType;
   bgColor?: types.BgColorType;
   style?: React.CSSProperties;
 };
@@ -25,6 +32,9 @@ export const Container = ({
   py,
   pt,
   pb,
+  mt,
+  mb,
+  my,
   bgColor,
   style,
   className,
@@ -39,6 +49,9 @@ export const Container = ({
       py ? paddingVerticals[py] : '',
       pt ? paddingTops[pt] : '',
       pb ? paddingBottoms[pb] : '',
+      my ? marginVerticals[my] : '',
+      mt ? marginTops[mt] : '',
+      mb ? marginBottoms[mb] : '',
       width ? styles.widths[width] : '',
       className,
     )}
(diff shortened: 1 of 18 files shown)


yvonnetangsu:
Gonna merge this in. @sherakama Please review retro if needed 🤠

components/Container/Container.tsx
```diff
@@ -15,6 +19,9 @@ export type ContainerProps = HTMLAttributes<HTMLElement> & {
   pt?: PaddingType;
   pb?: PaddingType;
   py?: PaddingType;
+  mt?: MarginType;
+  mb?: MarginType;
+  my?: MarginType;
```
sherakama, line 24: Curious to when a developer would use this parameter vs classname and what order of significance do they have (which one wins)?

components/StoryImage/StoryImage.tsx
```diff
@@ -0,0 +1,90 @@
...
+   * Crop width and height are used for width and height attributes on the img element.
+   * They don't need to be exact as long as the aspect ratio is correct.
+   */
+  const cropWidth = parseInt(cropSize?.split('x')[0], 10);
+  const cropHeight = aspectRatio === 'free'
+    ? Math.round(originalHeight * 2000 / originalWidth)
+    : parseInt(cropSize?.split('x')[1], 10);
+
+  return (
+    <WidthBox
```
sherakama, line 59: What are the differences, and rules, for when to use Container vs Grid vs Flexbox vs Widthbox? I am asking because it looks like you're setting up a convention here and it is probably something that should be documented for the next developer.

components/Container/Container.tsx
yvonnetangsu: I believe whichever classname the props is associated with appears later in the compiled CSS wins 😄 I added these mostly for convenience since it maps nicely to Storyblok datasource. I plan on creating a generic typography component also and will use them as well. It speeds things up for me when there are conditionals, eg, `py={!isMinimal ? 3 : undefined)` (this is for the padding ones)

components/StoryImage/StoryImage.tsx
yvonnetangsu: Good point - I'll document them when I come back.

You've seen the other ones so I'll talk about the `WidthBox` 😄 

So in Storyblok, I have several components set up in a way that a user can pick their bounding width and width (in number of columns) without nesting them inside a Section or Grid, e.g., WYSIWYG, Story Image, media embed. The WidthBox maps to these Storyblok options, eg, a user can add a Story Image, then pick a bounding width of "site" (which is the `cc`), then 10 of 12 columns - these two selected options are then passed into the WidthBox.

yvonnetangsu: I'd say the `WidthBox` is more useful for a headlessCMS like what we have in this case since it abstracts out some common layout options. If we're building a site without the CMS part then this won't add much.

PR #159, opened 2023-11-03
GIVCAMP-164: Add UTM params to outbound destinations.

sherakama:
# READY FOR REVIEW

# Summary
- Tracks inbound UTM codes in a cookie
- Adds UTM codes back to outbound links through the CTAExternalLink component.
- Adds filter to outbound URLs to match with stanford.edu TLD.

# Review By (Date)
- Whenever

# Criticality
-  Low

# Review Tasks

Basic Auth U/P: https://github.com/SU-SWS/ood-giving-campaign/blob/dev/netlify.toml#L29-L30

1. Review code
3. Try it out by navigating to the [preview URL with UTM params](https://deploy-preview-159--giving-campaign.netlify.app/?utm_source=test_source&utm_campaign=test_campaign&utm_medium=small&utm_term=test)
5. Click around to some internal pages and see the UTM params drop off the URL in the browser URL bar.
6. Click an outbound URL like the `Make a gift` button on the top right of the page (header)
7. Land on give.stanford.edu with the UTM params.


sherakama:
Removed review assignments as it seems to be full loading each page on Netlify at this time...

sherakama:
Ok, Assignments back in.

sherakama:
For context, the client does want to track the same campaign through multiple web properties and is managing them through a single GA account.

yvonnetangsu:
@sherakama This works well on Chrome, but it doesn't seem to preserve the utms on Safari desktop when clicking on external links 🤔

yvonnetangsu:
I'm no expert with dealing with utm params, but I did some really quick search and it looks like using cookies for utm params is a legit and fairly common approach 😄

sherakama:
> I'm no expert with dealing with utm params, but I did some really quick search and it looks like using cookies for utm params is a legit and fairly common approach 😄

Ok, but what about the other things, like, watching all the clicks and parsing the HTML element? I probably should check on mobile for this too.

sherakama:
@ericbakenhus thoughts and feedback?

components/GAProvider.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+      // Loop through the parent elements until an anchor element is found or parent becomes null
+      while (parentElement !== null) {
+        // Check if the current parent element is an anchor element
+        if (parentElement.tagName === 'A') {
+          return parentElement; // Found an anchor element, so return it
+        }
+
+        parentElement = parentElement.parentElement; // Move up to the next parent element
+      }
```
ericbakenhus, line 59: Might be able to use something like `target.closest('a')` if you're looking to find the nearest link ancestor (including self). Save a few lines of code.

https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

ericbakenhus:
> @ericbakenhus thoughts and feedback?

Quick thoughts:

- This won't capture links that are opened in a new tab
- This is adding the query variable to all absolute URLs, do we need an include/exclude list (e.g. the share links in individual articles)?
- We're definitely fine with the cookie setting the last UTMs for the entire session no matter what? For example, if I visit the site with UTM variables, then open the site again in a new tab, that original cookie will still be used for the new tab.

sherakama:
> This won't capture links that are opened in a new tab.

Oh, I always forget about the contextual click... hmm ok. Back to adding these to the anchor tag.

> This is adding the query variable to all absolute URLs, do we need an include/exclude list (e.g. the facebook/linkedin/twitter share links in individual articles)?

Yeah, good idea. I'll do a quick sani for outbound URLs that are XYZ.stanford.edu

> We're definitely fine with the cookie setting the last UTMs for the entire session no matter what? 

Yep.


Thanks @ericbakenhus

app/layout.tsx
```diff
@@ -55,30 +56,32 @@ storyblokInit({
...
-              <footer>
-                <LocalFooterMvp />
-                <GlobalFooter />
-              </footer>
-            </FlexBox>
-          </body>
-        </html>
-      </LazyMotionProvider>
-    </StoryblokProvider>
+    <GAProvider>
```
sherakama, line 59: Just adding the GAProvider

components/Cta/CtaExternalLink.tsx
```diff
@@ -24,12 +26,24 @@ export const CtaExternalLink = React.forwardRef<HTMLAnchorElement, CtaExternalLi
...
 
+    // Add UTM params to Stanford URLs.
+    const { isStanfordUrl, addUTMsToUrl } = useUTMs();
+    const [myHref, setMyHref] = useState<string>(href);
+    useEffect(() => {
+      if (typeof window === 'undefined') return;
+      if (isStanfordUrl(href)) {
+        setMyHref(addUTMsToUrl(href));
+      }
+    }, [href, isStanfordUrl, addUTMsToUrl]);
```
sherakama, line 41: Change the URL on the component mount if there is a cookie with UTMs available.

components/Cta/CtaLink.tsx
```diff
@@ -60,16 +59,12 @@ export const CtaLink = React.forwardRef<HTMLAnchorElement, CtaLinkProps>(
...
-    // Add UTM params to internal links
-    //const myLinkWithUtm = useAddUtmParams(myLink);
-    const myLinkWithUtm = myLink;
-
     if (isInternal) {
       return (
         <CtaNextLink
           {...rest}
           ref={ref}
-          href={myLinkWithUtm}
```
sherakama, line 72: Don't need em for internal urls

components/GAProvider.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+  useEffect(() => {
+    if (typeof window === 'undefined') {
+      return;
+    }
+    setUTMCookie();
+    return () => {
+      deleteUTMCookie();
+    };
+  }, [setUTMCookie, deleteUTMCookie]);
```
sherakama, line 16: On mount create a UTM storage cookie. On unmount, delete it.

sherakama:
Ok, I moved the altering of the URL into the CTAExternalLink component. That should account for the many ways someone can navigate off-site.

components/Cta/CtaLink.tsx
yvonnetangsu: Oh so the approach is different from Giving site, but this update makes more sense so we don't get ugly internal URLs 👍🏼

yvonnetangsu (approved):
Tested on Chrome, FF, Safari desktop + Chrome/Safari mobile and all worked for me 🎉 !

yvonnetangsu:
I see that you've addressed @ericbakenhus 's concern so I'm going to merge this in. Eric please feel free to comment after merge if you see anything else 😄

PR #176, opened 2023-12-15
Set slug route segment to revalidate 0

sherakama:
# READY FOR REVIEW

# Summary
- See if this helps with caching.


yvonnetangsu:
@sherakama So interesting, it is the revalidate to 0 that fixes the back button issue 😂 The back button in #177 is no longer working as we want it to, but this one still does. If only we can fix the error on those new pages we'll be good 😸

yvonnetangsu:
Actually, it looks like those pages are building after clearing cache 😄

yvonnetangsu (approved):
Weirdly this PR fixes the back button issue 😂 ! I'm going to approve and merge in so OOD can test on this version. I'll keep testing the caching/page error and will report if I see anything. Thank you again @sherakama ! 🎉

PR #183, opened 2023-12-21
GIVCAMP-225 | misc cleanup and sitemap update

yvonnetangsu:
# READY FOR REVIEW

# Summary
- First batch of misc updates
- Exclude pages with noindex=true in the sitemap

# Review By (Date)
- Soonish

# Review Tasks

1. Go to https://deploy-preview-183--giving-campaign.netlify.app/sitemap.xml
2. Confirm that `/page-not-found` isn't on the sitemap (I set it to noindex=true in Storyblok)
3. Go to https://deploy-preview-183--giving-campaign.netlify.app/stories/education-for-a-life-of-purpose
4. Inspect the code, check that in the hero, the superhead is set to aria-hidden=true
5. Check that the superhead is absorbed into the h1 as screen reader only text
Screenshot 2024-01-04 at 12 52 18 PM

- JIRA ticket(s) - GIVCAMP-225

### DIFF
--- a/app/sitemap.ts
+++ b/app/sitemap.ts
@@ -19,7 +19,10 @@ export default async function sitemap(): Promise<MetadataRoute.Sitemap> {
     cv: Date.now(),
   });
 
-  const ret = response.map((story) => {
+  // Exclude any stories with noindex set to true on Storyblok
+  const indexStories = response.filter((story) => !story.content?.noindex);
+
+  const ret = indexStories.map((story) => {
     return {
       url: story.path ?? `/${story.full_slug}`,
       lastModified: new Date(story.published_at),
--- a/components/Storyblok/SbTypeform.tsx
+++ b/components/Storyblok/SbTypeform.tsx
@@ -1,6 +1,6 @@
 import { storyblokEditable } from '@storyblok/react/rsc';
 import {
- PopOver, PopUp, SideTab, Slider, Widget,
+  PopOver, PopUp, SideTab, Slider, Widget,
 } from '@/components/TypeForm';
 import { Container } from '@/components/Container';
 import { type PaddingType } from '@/utilities/datasource';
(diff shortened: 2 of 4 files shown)


components/Storyblok/SbTypeform.tsx
```diff
@@ -32,7 +31,7 @@ export type SbTypeformProps = {
 };
 
 export const styles = {
-  container: 'cc flex flex-col place-items-center',
+  container: 'flex flex-col place-items-center',
```
yvonnetangsu, line 34: The default Container component `width="site"` setting adds the `cc` class, so I just remove both the `width="full"` below and the `cc` class here

app/sitemap.ts
```diff
@@ -19,7 +19,10 @@ export default async function sitemap(): Promise<MetadataRoute.Sitemap> {
     cv: Date.now(),
   });
 
-  const ret = response.map((story) => {
+  // Exclude any stories with noindex set to true on Storyblok
+  const indexStories = response.filter((story) => !story.content?.noindex);
```
yvonnetangsu, line 23: exclude any stories with noindex=true in Storyblok

sherakama (approved):
GTG

app/sitemap.ts
sherakama: The label in SB is a little misleading IMHO. It doesn't block search indexing but rather, removes it from the sitemap so the only way it would show up in google is by it being crawled. If you want to block the page from being indexed, you need to set a header or meta tag to tell google not to index that path.

yvonnetangsu: Hi @sherakama, I actually did add the meta tag for noindex previously in the getPageMetadata utility here:
https://github.com/SU-SWS/ood-giving-campaign/blob/04f9a3a3fd194d69a5791c6991fb2f6ba1acf524/utilities/getPageMetadata.ts#L99

Screenshot 2024-01-04 at 11 21 32 PM


In this PR, I decided to also exclude those noindex pages in the sitemap to avoid giving search engines conflicting info.

sherakama: Ah, that makes sense. Thank you for clarifying.

yvonnetangsu:
Thank you for the review, @sherakama! Going to merge this in 😄

PR #186, opened 2024-01-09
NoJira: Storyblok Editor

sherakama:
# Ready For Review

# Summary
- Fixes storyblok editor update issue

# Review By (Date)
- Soon

# Criticality
- Low

# Review Tasks

1. Navigate to the [storyblok editor](https://app.storyblok.com/#/me/spaces/1005200/stories/0/0/1479307/blok/029bb9a1-d087-4785-9d8c-b2b89ff53027?env=Debug)
2. Use the preview build as the environment (called debug in storyblok UI)
3. Edit some content and see the changes


components/StoryblokProvider.tsx
```diff
@@ -82,7 +82,7 @@ export default function StoryblokProvider({ children, isEditor = false }: Provid
   if (isEditor) {
     if (typeof window !== 'undefined') {
       const urlParams = new URLSearchParams(window.location.search);
-      accessToken = urlParams.get('token') || accessToken;
+      accessToken = urlParams.get('_storyblok_tk[token]') || accessToken;
```
sherakama, line 85: This changed...

sherakama:
@yvonnetangsu,

It looks like a package update (I suspect [@storyblok/react](https://github.com/storyblok/storyblok-react/compare/v2.4.7...v2.4.8)) caused the issue in the editor. 

I've added this deploy preview to the list of options in Storyblok if you want to test for yourself. 

I also moved a bunch of things around trying to rule stuff out. Some of the changes I think we can keep. I'll leave some comments.

I also see that there is a [new major version](https://github.com/storyblok/storyblok-react/releases/tag/v3.0.2).

I'll look into upgrading if we can with our versions.

app/(editor)/editor/layout.tsx
sherakama, line 1: Moved the editor to its own route group. More separation seems good here. I liked the file organization when it was tucked under the storyblok group better but this way there is more separation between layout.ts files.

app/(editor)/editor/page.tsx
```diff
@@ -6,8 +6,7 @@ import { components as Components } from '@/components/StoryblokProvider';
 import { resolveRelations } from '@/utilities/resolveRelations';
 import { notFound } from 'next/navigation';
 import ComponentNotFound from '@/components/Storyblok/ComponentNotFound';
-
-const activeEnv = process.env.NODE_ENV || 'development';
+import { ISbResult } from '@storyblok/react';
```
sherakama, line 9: Added type.

app/(editor)/editor/page.tsx
```diff
@@ -27,7 +26,6 @@ type PageProps = {
 
 // Control what happens when a dynamic segment is visited that was not generated with generateStaticParams.
 export const dynamic = 'force-dynamic';
-export const revalidate = 0;
```
sherakama, line 30: force-dynamic and revalidate = 0 do the same thing.

app/(editor)/editor/page.tsx
```diff
@@ -6,8 +6,7 @@ import { components as Components } from '@/components/StoryblokProvider';
 import { resolveRelations } from '@/utilities/resolveRelations';
 import { notFound } from 'next/navigation';
 import ComponentNotFound from '@/components/Storyblok/ComponentNotFound';
-
-const activeEnv = process.env.NODE_ENV || 'development';
```
sherakama, line 10: I got rid of this check. If the content is coming through the `(storyblok)` route segment assume only published content and vice versa for the `(editor)` route group. This seems like a safe assumption to me but let me know if you think otherwise.

app/(storyblok)/[[...slug]]/page.tsx
```diff
@@ -48,13 +48,11 @@ storyblokInit({
...
  */
 export async function generateStaticParams() {
-  const activeEnv = process.env.NODE_ENV || 'development';
 
   // Fetch new content from storyblok.
   const storyblokApi: StoryblokClient = getStoryblokApi();
   let sbParams: ISbStoriesParams = {
-    version: activeEnv === 'development' ? 'draft' : 'published',
-    cv: activeEnv === 'development' ? Date.now() : undefined,
+    version: 'published',
```
Comment (sherakama) [inline app/(storyblok)/[[...slug]]/page.tsx:55]: In the `(storyblok)` route group assume only published content.

public/icon1.png
sherakama, line 1: I moved these icon files to the public directory as I didn't see anywhere they were being imported into a component. Let me know if these are used and/or are still needed.

package.json
```diff
@@ -33,7 +33,7 @@
   },
   "dependencies": {
     "@heroicons/react": "^2.0.18",
-    "@storyblok/react": "^2.4.7",
+    "@storyblok/react": "^3.0.2",
```
sherakama, line 36: <sarcasm>My favorite kind of release notes. Major change. Warning warning. And nothing else... 😠 </sarcasm>

https://github.com/storyblok/storyblok-react/releases/tag/v3.0.0

app/(editor)/editor/page.tsx
yvonnetangsu: What if we're previewing locally but not in the visual editor iFrame? Eg, sometimes I like to go directly to localhost:3000 in my browser.

package.json
yvonnetangsu: Yikes...Storyblok 😬

public/icon1.png
yvonnetangsu: Oh these are actually for favicon/apple icons in the HTML head. Nextjs recommends that we add them this way in the `/app` directory as "metadata files". I believe they need to be in the root of `/app`
https://nextjs.org/docs/app/api-reference/file-conventions/metadata/app-icons

Currently on live:
Screenshot 2024-01-09 at 10 18 08 PM

Screenshot 2024-01-09 at 10 20 51 PM

yvonnetangsu:
@sherakama Is it possible to do the redirect from `/home` to `/` again (for the logo)?

yvonnetangsu:
Oof @sherakama I'm getting the back button issue in Safari again 😂

app/(editor)/editor/page.tsx
sherakama: That is a fair use case. I'll restore it.

sherakama:
@yvonnetangsu Things are back in their places. I am able to edit in storyblok and don't have the Safari bug.

yvonnetangsu:
> @yvonnetangsu Things are back in their places. I am able to edit in storyblok and don't have the Safari bug.

Woohoo! 😸

yvonnetangsu (approved):
Thanks for this fix, @sherakama! Let's merge it. I'm still seeing a Safari issue however, but it's slightly different this time. I've been able to reproduce on Safari desktop and mobile. Try this:

1. Go to the preview homepage
2. Click inside a story
3. Click the masthead logo to go back to the homepage
4. Now click the browser back button to go back to the story
5. Wait a few seconds - you're probably going to get redirected back to the homepage (you might have to wait a few seconds to see it)

I've found that this time the problem is only there if one clicks on the masthead logo first to navigate to the homepage, then click the back button. If the user never clicks any logo link (maybe something to do with the /home), then there's no issues with using the back button.

yvonnetangsu:
😂 Oh nextjs....

sherakama:
It's not nextjs exactly...

Something is up with the safari back button, next, and any javascript error. When any XHR or Fetch call fails from any of the resources being pre or regular fetched the browser is navigating back once more.

sherakama:
The home page thing was coincidental I think as the prefetch for that was causing a 500 error before I added the page.tsx index page. 

react-player was causing some fun so I updated that

Now it looks like some images may also be triggering...

I'm able to reproduce now with the browser console open at least.

sherakama:
@yvonnetangsu you ok to merge this to get the editor back working? I'll continue to fight the safari bug.

yvonnetangsu:
> @yvonnetangsu you ok to merge this to get the editor back working? I'll continue to fight the safari bug.

Yes, I'm ok with that! I tested again, and it looks like the Safari issue is still only confined to when people click the logo link and then use the back button, which is hopefully fairly rare. Thank you for all the investigation!

yvonnetangsu:
I'm going to push the button 😆

sherakama:
Great . Ty.

PR #191, opened 2024-01-12
GIVCAMP-226 | A11y better focus management for changemaker cards

yvonnetangsu:
# READY FOR REVIEW

# Summary
- A11y fix for updated changemaker cards
- Comment from SODA:

> Although focus no longer moves to the top of the page, the newly visible text is bypassed.  Move focus to the first content element in the card (the text).
> 
> Note: On Safari with VoiceOver the hidden card content (aria-hidden="true") is still announced.  However, this is a know bug that has not been resolved: https://bugs.webkit.org/show_bug.cgi?id=201887

# Review Tasks

1. Go to the changemaker cards
2. Use keyboard tab to focus on a card
3. Click enter
4. Check that the focus is moved to the content
5. Tab again and check that the focus is moved back to the "Dismiss" button
6. Confirm that by clicking enter that you can actually dismiss the content
7. Open another card content
8. Click escape and make sure it return focus to the button (it will have outline around the whole card)
9. Note: I'm not sure I need to address the Safari bug because if VO announces everything with aria-hidden then a lot of the things people normally do for a11y don't really work 😬 

- JIRA ticket(s) - GIVCAMP-226

### DIFF
--- a/components/ChangemakerCard/ChangemakerCard.styles.ts
+++ b/components/ChangemakerCard/ChangemakerCard.styles.ts
@@ -6,7 +6,7 @@ export const imageWrapper = 'overflow-hidden aspect-w-1 aspect-h-2';
 export const info = 'rs-px-1 pb-150 absolute w-full h-full bottom-0 left-0 mb-0';
 export const heading = 'mb-02em mt-auto';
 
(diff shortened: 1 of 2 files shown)


yvonnetangsu:
@sherakama This a11y update is ready for your eyes

sherakama (approved):
Works well for me. 

I think it was clear that SODA was saying that you don't need to worry about the Safari bug. GTG.

yvonnetangsu:
Thank you for the 👀 @sherakama ! Merging

PR #200, opened 2024-01-24
GIVCAMP-276: Start stubbing out the filters page.

sherakama:
# NOT READY FOR REVIEW
- (Edit the above to reflect status)

# Summary
- TL;DR - what's this PR for?

# Review By (Date)
- When does this need to be reviewed by?

# Criticality
- How critical is this PR on a 1-10 scale? Also see [Severity Assessment](https://stanfordits.atlassian.net/browse/D8CORE-1705).
- E.g., it affects one site, or every site and product?

# Review Tasks

1. Check out this branch or go to the netlify preview
2. Navigate to...
3. Verify...

- JIRA ticket(s) - GIVCAMP-
- Other PRs
- Any other contextual information that might be helpful (e.g., description of a bug that this PR fixes, new functionality that it adds, etc.)
- Anyone who should be notified? (`@mention` them here)


app/(storyblok)/loading.tsx
```diff
@@ -11,11 +10,7 @@ const Loading = () => {
       <main>
         <Container width='site' className='rs-my-8'>
           <FlexBox gap direction='row' wrap='wrap'>
-            <Skeleton height={300} highlightColor='rgba(67,66,62,.8)' containerClassName='w-full'/>
-            <Skeleton height={300} highlightColor='rgba(67,66,62,.8)' containerClassName='min-w-[25%] flex-grow' />
-            <Skeleton height={300} highlightColor='rgba(67,66,62,.8)' containerClassName='min-w-[25%] flex-grow' />
-            <Skeleton height={300} highlightColor='rgba(67,66,62,.8)' containerClassName='min-w-[25%] flex-grow' />
-            <Skeleton height={200} highlightColor='rgba(67,66,62,.8)' containerClassName='min-w-[40%] flex-grow' />
+            <Rectangle />
```
yvonnetangsu, line 13: Oh I'm tweaking it in my other PR too 🤣

sherakama:
Been banging on this today. 

Works on local just fine.

On Netlify, when router.push() happens, there is a fetch request that goes out that is supposed to respond with the RSC payload of filtered results. However, it doesn't seem to be hitting the functions as nothing gets logged, and the default (All results) payload is returned. Something somewhere is cached or not being redirected properly on the Netlify infrastructure. 

I might just punt on the server side rendering anyhow as we want isotope/masonry style filtering and that is client-side work.

sherakama:
Ok, well https://github.com/SU-SWS/ood-giving-campaign/pull/200/commits/1c52ab5076ed49eab68ac18644a20a020f85e59b confirms it. Something is up with the Nextlify plugin when using the [slug] route at the top level and trying to add another route `/taxonomy`. The direct routing works with the search params eg: https://deploy-preview-200--giving-campaign.netlify.app/taxonomy?schools=doerr but if you try to soft navigate using `router.push` the fetch request for the RSC payload returns only the cached `/taxonomy` route without the search params even if the search params are present in the request.

sherakama:
I'm going to make the taxonomy page a Storyblok page and the filter and results as components. Let's see if we can get that to work as intended.

sherakama:
This request for med returned back the full list. 
Screenshot 2024-01-25 at 09 24 39

yvonnetangsu:
> This request for med returned back the full list. Screenshot 2024-01-25 at 09 24 39

Very nice!

yvonnetangsu:
> I'm going to make the taxonomy page a Storyblok page and the filter and results as components. Let's see if we can get that to work as intended.

That would be perfect because there're probably going to be curated elements above and below the filtered area 👍🏼 😄

sherakama:
@yvonnetangsu 

Got the bones for this in place now https://deploy-preview-200--giving-campaign.netlify.app/test/taxonomy

This is all client-side fetching and is set up to fetch everything at the moment. I haven't looked into the 'view more' functionality but it gives us the gist of how it should be built out. 

The TLDR of this thread is that there is a conflict with the Netlify Plugin and the catch all [slug] route that was causing errors when trying to do this as SSR and useSearchParams(). I switched to API routes and client side fetching with a context wrapper for the two components. The two components are filters and results and they can be moved around in Storyblok. 

I'll hold on this now until the design settles.

yvonnetangsu:
Thank you for doing this exploration @sherakama! This would be super useful when we get to phase. Looks like we're going with a much simpler approach at the moment, but eventually we're going to need this. Happy Friday 🚀

sherakama:
Should probably look at coming back to this with plugin 5.0.0-beta.x

PR #212, opened 2024-01-30
Update cron.yml

sherakama:
# RETRO REVIEW

# Summary
- I forgot this is a public repo and we really shouldn't publish these webooks.
- I have removed the webhook from the code repo and have revoked it on Netlify.
- The new one has been added as a secret
- I'll update Storyblok as well.


yvonnetangsu:
Thanks!

PR #237, opened 2024-02-28
NoJira: Remove Editor Token

sherakama:
# READY FOR REVIEW

# Summary
- Per client request: We really only share them with internal team members and leadership, but do rely on them working/being available for many days, sometimes weeks, for review – particularly by leadership.

This opens up the URL and editor to live for a very long time. The risk associated with this is the possibility of someone getting a hold of the API token and using it against the preview URL to see unpublished content. This is a low-risk item to me as they could also use the same API key to query the API directly to get the same content. The preview URL is just a convenient way. One could also just change the timestamp if they knew.

# Review By (Date)
- End of the week?

# Criticality
- Normal

# Review Tasks

1. Review code
2. Add Preview URL to Storyblok space
3. View content in Storyblok editor
4. Click on the preview button in the Storyblok editor to open up a full page preview
5. Save the URL somewhere and come back 4 hours later to see if it still works

### DIFF
--- a/app/(editor)/editor/page.tsx
+++ b/app/(editor)/editor/page.tsx
@@ -89,17 +89,14 @@ async function getStoryData({ path }: PageProps['searchParams']): Promise<ISbRes
 /**
  * Validate the editor token.
  *
- * We expect 3 hours as the time of a token being valid.
- * So basically that an editor will work max 3 hours (10800 seconds) on one page
- * before switching to another entry inside the editor or refreshing the browser window.
- * You can extend that by adjusting 10800 with the value you need.
+ * Removed time limit check to support client workflows of several days, or weeks 
+ * of using the preview link for review.
  */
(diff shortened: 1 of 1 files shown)


yvonnetangsu:
Thank you for this @sherakama I'll review tomorrow. I think Alexis would be excited about this! 😄 🎉

yvonnetangsu:
@sherakama I don't think it's this PR, but I'm seeing a strange behavior. If I go to a Storyblok page that isn't the homepage, when I select one of the deploy preview URLs from the visual editor, I always see the homepage showing up instead of that page's content. If I select the production/dev/local URL then there is no problem. I remember that the deploy preview URLs worked before - not sure what changes we made would cause that. Or am I doing something weird? 🤔

yvonnetangsu (approved):
PR GTG! Thank you for this @sherakama ! 🚀

PR #269, opened 2024-04-16
GIVCAMP-298 | Row component with spacers; misc updates

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Row and Spacer component
- Misc updates (see things called out below)

# Review By (Date)
- Retro

# Criticality
- 3

# Review Tasks

1. Go to storyblok test/row-test
2. Pick PR 269 from the visual editor URL
3. Look at all the examples and the options available

- JIRA ticket(s) - GIVCAMP-298

### DIFF
--- a/components/Sidebar/SidebarCard.styles.ts
+++ b/components/Sidebar/SidebarCard.styles.ts
@@ -1,12 +1,13 @@
 import { cnb } from 'cnbuilder';
 
-export const root = (hasBgColor?: boolean) => cnb('relative overflow-hidden @container', hasBgColor ? 'text-gc-black backdrop-blur-sm' : '');
+export const root = (hasBgColor: boolean) => cnb('relative overflow-hidden @container', hasBgColor ? 'text-gc-black backdrop-blur-sm' : '');
 
 export const content = (
-  hasBgColor?: boolean,
-  hasBarColor?: boolean,
-  barOnRight?: boolean,
-) => cnb('', {
+  isLightText: boolean,
+  hasBgColor: boolean,
+  hasBarColor: boolean,
+  barOnRight: boolean,
+) => cnb(isLightText ? 'text-white' : '', {
   'rs-py-1': hasBgColor,
   'rs-py-2': !hasBgColor,
   'pr-0': !hasBgColor && hasBarColor && !barOnRight,
--- a/components/Storyblok/SbText.tsx
+++ b/components/Storyblok/SbText.tsx
@@ -27,6 +27,7 @@ export type SbTextProps = {
     align?: TextAlignType;
     weight?: FontWeightType;
     italic?: boolean;
+    uppercase?: boolean;
     srOnly?: boolean;
     boundingWidth?: 'site' | 'full';
     width?: '12' | '10' | '8' | '6' | '4';
@@ -51,6 +52,7 @@ export const SbText = ({
     align,
     weight,
     italic,
+    uppercase,
     srOnly,
     boundingWidth,
     marginTop,
@@ -82,6 +84,7 @@ export const SbText = ({
         align={align}
         weight={weight}
         italic={italic}
+        uppercase={uppercase}
         srOnly={srOnly}
         className="whitespace-pre-line mb-0"
       >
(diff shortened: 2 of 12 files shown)


components/Sidebar/SidebarCard.styles.ts
```diff
@@ -1,12 +1,13 @@
...
 
-export const root = (hasBgColor?: boolean) => cnb('relative overflow-hidden @container', hasBgColor ? 'text-gc-black backdrop-blur-sm' : '');
+export const root = (hasBgColor: boolean) => cnb('relative overflow-hidden @container', hasBgColor ? 'text-gc-black backdrop-blur-sm' : '');
 
 export const content = (
-  hasBgColor?: boolean,
-  hasBarColor?: boolean,
-  barOnRight?: boolean,
-) => cnb('', {
+  isLightText: boolean,
```
yvonnetangsu, line 6: Added a light text option for sidebar card

components/TextCard/TextCard.styles.ts
```diff
@@ -1,4 +1,4 @@
-export const root = 'relative rs-py-5 rs-px-3';
+export const root = 'relative rs-py-5 rs-px-3 break-words';
```
yvonnetangsu, line 1: Allow long words to break into 2nd row in TextCard

components/Storyblok/SbText.tsx
```diff
@@ -27,6 +27,7 @@ export type SbTextProps = {
     align?: TextAlignType;
     weight?: FontWeightType;
     italic?: boolean;
+    uppercase?: boolean;
```
yvonnetangsu, line 30: Add uppercase option for SbText (Typography)

sherakama:
Those are a lot of options. 

This all seems to work. A couple of notes for you

1. If the 'Make a row an unordered list' short normally be checked, please default it to enabled and allow the content author to turn off when they add a spacer
2. This looks like another powertool with tonnes of combinations and when used within sections or other containers can create a lot of places for spacing or style changes can happen. I have concerns that it will lead to some difficulty for the content author to troubleshoot or get set 'just right' when spacing can be applied at multiple layers.

sherakama (approved):

yvonnetangsu:
> Those are a lot of options.
> 
> This all seems to work. A couple of notes for you
> 
> 1. If the 'Make a row an unordered list' short normally be checked, please default it to enabled and allow the content author to turn off when they add a spacer
> 2. This looks like another powertool with tonnes of combinations and when used within sections or other containers can create a lot of places for spacing or style changes to happen. I have concerns that it will lead to some difficulty for the content author to troubleshoot or get set 'just right' when spacing can be applied at multiple layers.

Thank you for the review @sherakama ! Yup lots of options - the OOD team has earned it 😆 For 1), I have the list option default to false because I think it would most commonly used for having WYSIWYG and image side by side. For Story Cards, they usually use the Grid component so I'm not too worried about this one. 2) Yes you're right! With that many options sometimes things can get tricky. I'll create a bunch of presets for them to mitigate potential problems 👍🏼

sherakama:
> Yup lots of options - the OOD team has earned it 

They've definitely proven themselves as one of our most sophisticated content author groups.

> I have the list option default to false because I think it would most commonly be used for having WYSIWYG and image side by side. 

Ok, something to keep an eye on in the Siteimprove reports. 

> Yes you're right! With that many options sometimes things can get tricky. I'll create a bunch of presets for them to mitigate potential problems 👍🏼

Thanks, the presets are definitely helpful. We have a savvy bunch of content authors in this group so I am less worried but I thought I would chime in anyhow as something to think about for the next group.

yvonnetangsu:
> > Yup lots of options - the OOD team has earned it
> 
> They've definitely proven themselves as one of our most sophisticated content author groups.
> 
> > I have the list option default to false because I think it would most commonly be used for having WYSIWYG and image side by side.
> 
> Ok, something to keep an eye on in the Siteimprove reports.
> 
> > Yes you're right! With that many options sometimes things can get tricky. I'll create a bunch of presets for them to mitigate potential problems 👍🏼
> 
> Thanks, the presets are definitely helpful. We have a savvy bunch of content authors in this group so I am less worried but I thought I would chime in anyhow as something to think about for the next group.

Yup yup, those are very good points @sherakama ! 😄  If I'm working with authors who are less experienced with Storyblok I'll mostly likely hold back on adding that many options 🤠

yvonnetangsu:
Merging this into `dev` and will demo it at today's meeting 🤠

PR #273, opened 2024-04-27
GIVCAMP-342 | Add taxonomy chips and light/dark mode to all variants of Story Cards

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Add taxonomy chips (initiative only for now) to all variants of Story Cards
- Add light/dark mode to Story Cards (it was inheriting parent container text color currently so it still sort of work, but this actual light/dark mode selection will modify the variant of the chips also (light/dark)

# Review By (Date)
- Retro

# Criticality
- 6

# Review Tasks

1. Go to Storyblok `test/home-new-test` and select PR 273 from the visual editor URL
2. See the chips on the horizontal, list, and default (vertical) story card variants
3. Test responsive behavior and see that all chips are left aligned with the card body copy.
4. Go to the newsletter page and look at the light mode cards with the light mode tags. NOTE: Once this is live, we'll change the grid behavior of some of these card grids since the story cards are now too small when they have tags added.
Screenshot 2024-05-01 at 4 55 30 PM

- JIRA ticket(s) - GIVCAMP-342

### DIFF
--- a/components/Cta/Cta.styles.ts
+++ b/components/Cta/Cta.styles.ts
@@ -20,7 +20,7 @@ const brochureIlluminating = 'from-illuminating to-illuminating';
 const brochurePoppy = 'from-poppy to-poppy';
 
 const chipBase = 'relative inline-block border leading-display rounded-full font-normal no-underline underline-offset-4 hocus:underline hocus:bg-digital-red-light hocus:text-white';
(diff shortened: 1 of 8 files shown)


yvonnetangsu:
Merging this into the integration branch 🤠

PR #283, opened 2024-05-09
Nojira | Use Campaign black for body text color in TW base

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Use Campaign black as base HTML body text color

# Review By (Date)
- Retro

# Criticality
- 2

# Review Tasks

1. I've asked the client to QA this and it has passed so we're good.

### DIFF
--- a/tailwind/plugins/base/gc-base.js
+++ b/tailwind/plugins/base/gc-base.js
@@ -11,6 +11,7 @@ module.exports = function () {
       },
       body: {
         fontSize: '1.8rem',
+        color: '#17171A',
 
         '@screen md': {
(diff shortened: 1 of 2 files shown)


tailwind/plugins/base/gc-base.js
```diff
@@ -11,6 +11,7 @@ module.exports = function () {
       },
       body: {
         fontSize: '1.8rem',
+        color: '#17171A',
```
yvonnetangsu, line 14: This makes sure that the fallback color for text everywhere is this Campaign Black and not the softer Decanter black.

PR #290, opened 2024-05-14
GIVCAMP-351 | Add themes to taxonomy in story cards

yvonnetangsu:
# READY FOR REVIEW

# Summary
- This gets merged into the integration branch `story-list-pages` after approval
- Add themes to the taxonomy tags
- Content authors will self governed (decided in meeting) to not tag more than 3 items and they will order them in Storyblok
- Initiative tags go before any theme tags

# Review By (Date)
- Retro

# Criticality
- 5

# Review Tasks

1. Go to https://deploy-preview-290--giving-campaign.netlify.app/ and see the story tags on the story cards.
1. Go to Test -> Home New Test and select PR290 from the visual editor in Storyblok
2. Play around with taxonomies (initiatives and Themes only). Check that they're displayed with initiative tags first, followed by theme tags
3. For a story card, select any heading level except h6, check that the screen reader only tag heading "Story tags:"  level is one level down. Select heading level h6, check code that the tag heading has heading level h6 because h7 doesn't exist.
4. Look at code and things called out

- JIRA ticket(s) - GIVCAMP-351

### DIFF
--- a/components/StoryCard/StoryCard.tsx
+++ b/components/StoryCard/StoryCard.tsx
@@ -8,7 +8,7 @@ import {
 import { SbLinkType } from '@/components/Storyblok/Storyblok.types';
 import { getProcessedImage } from '@/utilities/getProcessedImage';
 import { accentBorderColors, type AccentBorderColorType } from '@/utilities/datasource';
-import { initiativesMap, type InitiativesType } from '@/utilities/taxonomyMaps';
+import { taxonomyMap, type TaxonomyType } from '@/utilities/taxonomyMaps';
 import * as styles from './StoryCard.styles';
 
 export type StoryCardProps = React.HTMLAttributes<HTMLDivElement> & {
@@ -21,7 +21,7 @@ export type StoryCardProps = React.HTMLAttributes<HTMLDivElement> & {
   tabColor?: AccentBorderColorType;
   href?: string;
   link?: SbLinkType;
-  taxonomy?: InitiativesType[];
+  taxonomy?: TaxonomyType[];
   animation?: AnimationType;
   delay?: number;
   isHorizontal?: boolean;
@@ -124,7 +124,7 @@ export const StoryCard = ({
                 {taxonomy.slice(0, 3).map((item) => (
                   <li key={item} className={styles.taxonomyItem}>
                     <CtaLink href={`/stories/list/${item}`} variant={isDark ? 'storyCardChipDark' : 'storyCardChip'} className={styles.taxonomyLink}>
-                      {initiativesMap[item]}
+                      {taxonomyMap[item]}
                     </CtaLink>
                   </li>
                 ))}
--- a/components/StoryCard/StoryCard.tsx
+++ b/components/StoryCard/StoryCard.tsx
@@ -3,8 +3,8 @@ import { AnimateInView, type AnimationType } from '@/components/Animate';
 import { CtaLink } from '@/components/Cta/CtaLink';
 import { FlexBox } from '@/components/FlexBox';
 import {
-  Heading, type HeadingType, Paragraph, type FontSizeType,
-} from '../Typography';
+  Heading, Paragraph, type HeadingType, type HeadingLevelNumberType,
+} from '@/components/Typography';
 import { SbLinkType } from '@/components/Storyblok/Storyblok.types';
 import { getProcessedImage } from '@/utilities/getProcessedImage';
 import { accentBorderColors, type AccentBorderColorType } from '@/utilities/datasource';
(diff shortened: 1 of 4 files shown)


components/Typography/typography.types.ts
```diff
@@ -1,6 +1,7 @@
 import * as styles from './typography.styles';
 
-export type HeadingType = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+export type HeadingLevelNumberType = 1 | 2 | 3 | 4 | 5 | 6;
+export type HeadingType = `h${HeadingLevelNumberType}`;
```
yvonnetangsu, line 4: Do this so I can make the tag headling level type a bit more robust

utilities/taxonomyMaps.ts
```diff
@@ -1,7 +1,7 @@
 export const initiativesMap = {
   'ethics-society-and-technology': 'Ethics, Society & Technology',
   'human-centered-artificial-intelligence': 'Human-Centered Artificial Intelligence',
-  'innovative-medicines-accelerator': 'innovative-medicines-accelerator',
+  'innovative-medicines-accelerator': 'Innovative Medicines Accelerator',
```
yvonnetangsu, line 4: I blame it on copilot here 👀

utilities/taxonomyMaps.ts
```diff
@@ -15,3 +15,17 @@ export const initiativesMap = {
...
+  'preparing-citizens': 'Preparing Citizens',
+  'sustaining-life': 'Sustaining Life',
+};
+export type ThemesType = keyof typeof themesMap;
+
+export const taxonomyMap = {
+  ...initiativesMap,
+  ...themesMap,
+};
+export type TaxonomyType = keyof typeof taxonomyMap;
```
yvonnetangsu, line 31: Taxonomy array is just the combination of the initiative and the themes one

components/StoryCard/StoryCard.tsx
```diff
@@ -48,12 +48,11 @@ export const StoryCard = ({
...
-  if (isHorizontal) {
-    headingSize = 'f5';
-  } else if (isSmallHeading && !isHorizontal) {
-    headingSize = 2;
-  };
+  // The heading level of the tags should be one more than the heading level of the card, but maxes out at h6
+  const tagsHeadingLevelNumber: HeadingLevelNumberType =
+    Math.min(parseInt(headingLevel?.slice(1), 10) + 1, 6) as HeadingLevelNumberType;
+
+  const tagsHeadingLevel: HeadingType = `h${tagsHeadingLevelNumber}`;
```
yvonnetangsu, line 55: Add a screen reader heading for the story tags, which is one level down compared to the heading level of the Card heading, but it needs to max out at h6 (which is unlikely scenario anyway that we'll actually use a h6 for the card heading)

I'm trying my best with TypeScript but there is no way for me to get around not using the as HeadingLevelNumberType here. I think this is robust enough for this project though.

sherakama: I was looking to see if we can set an Integer Range type and the best answer I found was here: https://stackoverflow.com/questions/39494689/is-it-possible-to-restrict-number-to-a-certain-range

sherakama: I guess the refactor here would be to change the `headingLevel` prop from a set of strings to the integer range of 1 - 6 and then to construct the heading level string to pass down into the `<Heading>` component.

utilities/taxonomyMaps.ts
```diff
@@ -15,3 +15,17 @@ export const initiativesMap = {
   'undergraduate-financial-aid': 'Undergraduate Financial Aid',
 };
 export type InitiativesType = keyof typeof initiativesMap;
+
+export const themesMap = {
+  'accelerating-solutions': 'Accelerating Solutions',
+  'catalyzing-discovery': 'Catalyzing Discovery',
+  'preparing-citizens': 'Preparing Citizens',
+  'sustaining-life': 'Sustaining Life',
+};
```
sherakama, line 24: So all of these maps are copies of the key/label pairs in the datasource config?

I'm not crazy about duplicating and hard coding these values here as ideally they would/could be edited like content in the CMS. If we cannot pull and infer the types I'd rather see these set to string.

sherakama: `{[key:string]: string}[]`

sherakama: https://bobbyhadz.com/blog/typescript-key-string-string

yvonnetangsu: Yeah I can see pulling from datasource might be better. Let me look into server side fetching for these 🤔

components/StoryCard/StoryCard.tsx
yvonnetangsu: The range restriction is indeed cool 😄 , but, I might cheat a bit for this heading level here since for what we use it for (or just holding off for a bit till we're done with the story nav feature), it would be very unlikely that we run into any issues.

sherakama: Yeah, that is totally fine here.

utilities/taxonomyMaps.ts
yvonnetangsu: @sherakama - I'm going to add a TODO here for fetching server side after I come back. I'll create a ticket so it won't be an empty promise 😁 After that I'm going to merge this branch into the integration branch just so I can keep rolling things in before I go on vacay.

yvonnetangsu: Ticket created here:
https://stanfordits.atlassian.net/browse/GIVCAMP-355

yvonnetangsu:
Rolling into the integration branch `story-list-pages` with the TODO GIVCAMP-355. Thanks for keeping me honest @sherakama !

8qxhn3

PR #286, opened 2024-05-10
GIVCAMP-303 | Horizontal changemaker card; use modal for XS for all changemaker variants

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Horizontal variant of changemaker card
- For all variants of changemaker cards, use modal for content for only the XS breakpoint.
- Add Headless UI package

# Review By (Date)
- Retro

# Criticality
- 6

# Review Tasks

## Setup tasks and/or behavior to test

1. Go to Storyblok Test -> Yvonne Test Horizontal Changemakers and select the PR286 URL from the visual editor
2. Look at both the horizontal and vertical changemaker cards at all breakpoints.
3. Test on the phone and make sure the modal works
4. Checks code for a11y
5. Test on Firefox (see call out below for the bug and the solution)

# Associated Issues and/or People
- JIRA ticket(s) - GIVCAMP-303

PR #291, opened 2024-05-16
Release/v2.1.0

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Add HeadlessUI package (we need a modal for the changemaker cards)
- Add horizontal changemaker card variant
- Use modal for XS for all changemaker cards
- Add @netlify/plugin-nextjs back to the repo so we're using a known tested version

# Review By (Date)
- Retro

# Criticality
- 5

# Review Tasks

1. Look at the homepage and test that the changemaker cards are working fine. 
2. Changes should already be reviewed so we should be ok.

### DIFF
--- a/tailwind/plugins/base/gc-base.js
+++ b/tailwind/plugins/base/gc-base.js
@@ -6,7 +6,6 @@ module.exports = function () {
   return function ({ addBase, config }) {
     addBase({
       html: {
-        overflowY: 'visible !important', // Need this for sticky nav to work
         color: '#17171A', // Momentum black
       },
       body: {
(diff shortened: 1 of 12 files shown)


tailwind/plugins/base/gc-base.js
```diff
@@ -6,7 +6,6 @@ module.exports = function () {
   return function ({ addBase, config }) {
     addBase({
       html: {
-        overflowY: 'visible !important', // Need this for sticky nav to work
```
yvonnetangsu, line 9: This was for Gatsby and for Travel Study - we don't need it here. It interferes with how the body scroll locking work for the Headless UI mobile.

yvonnetangsu:
@sherakama I'm going to push this through as you've already reviewed the PR #286 Adding the netlify runtime plugin here as we discussed 🤠

PR #294, opened 2024-05-24
Nojira | Fixed nested routes in main nav by using SbLink directly rather than href

yvonnetangsu:
# READY FOR REVIEW
- (Edit the above to reflect status)

# Summary
- Fix wrong route due to missing `/` in front in main on pages that are nested 2 levels deep, e.g., on the Newsletter sign up page https://momentum.stanford.edu/newsletter/sign-up, if you click on the `Changemakers` link in the main nav, it will actually bring you to https://momentum.stanford.edu/newsletters/changemakers because the `href` is missing the `/` in front of the slug `changmakers` 🤦🏼‍♀️ 
- This is an easy fix because if I pass in the SbLink directly (which the `CtaLink` component takes - it takes either `href` or `SbLink`)
- This also means that the function `transformNavItems` can be removed so we got rid of it in this PR

# Review By (Date)
- Retro

# Criticality
- 8

# Review Tasks

1. Go to https://deploy-preview-294--giving-campaign.netlify.app/newsletter/sign-up
2. Open the main nav and click on Changemakers link, check that it goes to the Changemakers page, and not a 404 due to it trying to go to newsletters/changemakers


yvonnetangsu:
@sherakama Pushing this through with a release before anyone notices 😂 I have a good idea of what was wrong and have tested thoroughly this time. Please double check when you have a chance.

components/MainNav/MainNav.tsx
```diff
@@ -53,29 +48,31 @@ export const MainNav = ({
...
-                          </CtaLink>
-                        </li>
-                      ))}
-                    </ul>
+                    {!!featuredLinks?.length && (
+                      <ul className={styles.linkList}>
+                        {featuredLinks.map((link) => (
+                          <li key={link._uid} className={styles.featuredItem}>
+                            <CtaLink
+                              sbLink={link.link}
```
yvonnetangsu, line 56: I previously set up CtaLink to take both SbLink directly, or a href (so it can be used with Storyblok component directly as well as hardcoded with routes). I forgot I supports SbLink since it's been a while since I use it. When you pass a SbLink directly into CtaLink, it automatically add a / in front of the cached_url which is what we need.

PR #301, opened 2024-06-12
GIVCAMP-358 | Add filtered story lists

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Fetch stories to display in list view in story filter pages - top level at /stories (display all stories) and the filtered ones inside the directory /stories/list/
- Add skip link to skip past story filter menu on story filter pages (will work on mobile etc in another ticket)
- Allow the Storyblok CTA to have spacing top and bottom options

# Review By (Date)
- Soon would be nice

# Criticality
- 7

# Review Tasks

1. Go to Storyblok /stories page
2. Choose the PR 301 from the visual editor URL list
3. Use the tab key on the keyboard and see how the skip link appears for the story filter nav
Screenshot 2024-06-13 at 9 54 05 AM

4. Verify that the story list below the featured story (the large horizontal card) is displaying all published stories except for the featured story
5. Check that the stories are in reverse chronological order (featured story not part of this as it's manually added)
6. Try deleting the featured story card, verify that the featured story is now showing up in the list view
7. Go to /stories/list folder, click on the Preparing Citizens page
8. Check the same things as for the /stories page, except that now the list should only display stories that are tagged with Preparing Citizens.
9. Look at code for approach

- JIRA ticket(s) - GIVCAMP-358

### DIFF
--- a/app/(editor)/editor/page.tsx
+++ b/app/(editor)/editor/page.tsx
@@ -86,6 +86,44 @@ async function getStoryData({ path }: PageProps['searchParams']): Promise<ISbRes
 
 };
 
+// Get a list of stories that are of component sbStoryMvp
+async function getStories() {
+  const activeEnv = process.env.NODE_ENV || 'development';
+  const storyblokApi: StoryblokClient = getStoryblokApi();
+  const sbParams: ISbStoriesParams = {
+    version: 'published',
+    cv: activeEnv === 'development' ? Date.now() : undefined,
+    resolve_relations: resolveRelations,
+    'starts_with': 'stories/',
+    'sort_by': 'first_published_at:desc',
+    filter_query: {
+      component: {
+        in: 'sbStoryMvp',
+      },
+    },
+  };
+
+  try {
+    const stories = await storyblokApi.getAll('cdn/stories', sbParams);
+    return stories;
+  }
+  catch (error) {
+    if (typeof error === 'string') {
+      try {
+        const parsedError = JSON.parse(error);
+        if (parsedError.status === 404) {
+          return { data: 404 };
+        }
+      }
+      catch (e) {
+        console.error('Error', error);
+      }
+    }
+  }
+
+  return [];
+}
+
 /**
  * Validate the editor token.
  *
@@ -117,6 +155,7 @@ export default async function Page({ searchParams }: PageProps) {
 
   // Get data out of the API.
   const { data } = await getStoryData(searchParams);
+  const stories = await getStories();
 
   // Failed to fetch from API because story slug was not found.
   if (data === 404) {
@@ -125,6 +164,12 @@ export default async function Page({ searchParams }: PageProps) {
 
   // Return the story.
   return (
-    <StoryblokStory story={data.story} bridgeOptions={bridgeOptions} slug={slug} name={data.story.name} />
+    <StoryblokStory
+      story={data.story}
+      stories={stories}
+      bridgeOptions={bridgeOptions}
+      slug={slug}
+      name={data.story.name}
+    />
   );
 };
--- a/app/(editor)/editor/page.tsx
+++ b/app/(editor)/editor/page.tsx
@@ -98,15 +98,22 @@ async function getStoryData({ path }: PageProps['searchParams']): Promise<ISbRes
 /**
  * Get a list of stories that are of component sbStoryMvp in reverse chronological order.
  */
-async function getStories({ path }: PageProps['searchParams']) {
+async function getStoryList({ path }: PageProps['searchParams']) {
   const activeEnv = process.env.NODE_ENV || 'development';
   const storyblokApi: StoryblokClient = getStoryblokApi();
   const fullslug = path.replace(/\/$/, '') || 'home';
 
   // Get the last part of the path.
   const slug = path.split('/').pop() || '';
-  // let slug = '';
+
   let orQuery: FilterQuery[] = [];
+
+  /**
+   * If the page is inside the folder stories/list/ (story list pages filtered by taxonomy),
+   * add a filter query to only return stories that has an initiative or theme that matches the slug of that story.
+   * E.g., if the full slug is stories/list/preparing-citizens,
+   * only return stories that have 'preparing-citizens' tagged as a theme or initiative.
+   */
   if (fullslug.includes('stories/list/')) {
     orQuery = [
       {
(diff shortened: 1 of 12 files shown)


components/Hero/StoryListHero.tsx
```diff
@@ -35,10 +35,10 @@ export const StoryListHero = ({
             align="center"
             leading="none"
             color="white"
-            className={styles.heading}
+            className={styles.heading(!!subheading)}
           >
             <Text as="span" font="druk" size={subheading ? 'f5' : 'splash'} leading="druk">Stories</Text>
-            <Text as="span" font="serif" size={subheading ? 'f4' : 'f7'} weight="normal" italic className="relative -top-[.23em] -ml-01em mr-02em"> of </Text>
+            <Text as="span" font="serif" size={subheading ? 'f4' : 'f7'} weight="normal" italic className="relative -top-[.23em] -ml-[.07em] mr-01em"> of </Text>
```
yvonnetangsu, line 41: I'll move these over to the styles file in the mobile styles PR 😂

utilities/taxonomyMaps.ts
```diff
@@ -2,7 +2,7 @@
 // Per PR discussion with Shea https://github.com/SU-SWS/ood-giving-campaign/pull/290#discussion_r1602386147
```
yvonnetangsu, line 2: I'll work on this after the mobile styles (fetching the datasource for the taxonomy)

components/Storyblok/SbStoryFilterPage/SbStoryFilterPage.tsx
```diff
@@ -35,18 +51,86 @@ export const SbStoryFilterPage = ({
...
+   * which is more efficient than using an array for the same purpose.
+   */
+  const featuredStoryUUIDSet = new Set(featuredStories?.map(item => item.storyPicker.uuid));
+
+  /**
+   * Filter storyList (list of all stories with taxonomy that matches the slug of the page)
+   * to exclude items with uuid that are in the featuredStoryUUIDArray,
+   * ie, we remove stories that are already added as featured stories cards.
+   */
+  const filteredStoryList = storyList.filter(item => !featuredStoryUUIDSet.has(item.uuid));
```
yvonnetangsu, line 68: This is how I exclude the featured stories from the stories rendered in the list.

app/(editor)/editor/page.tsx
```diff
@@ -86,6 +95,65 @@ async function getStoryData({ path }: PageProps['searchParams']): Promise<ISbRes
...
+    version: 'published',
+    cv: activeEnv === 'development' ? Date.now() : undefined,
+    starts_with: 'stories/',
+    sort_by: 'first_published_at:desc',
+    per_page: 100,
+    filter_query: {
+      component: {
+        in: 'sbStoryMvp',
+      },
+      __or: orQuery,
```
yvonnetangsu, line 143: `__or` is how we can add query using the or operator. So here either the slug matches an item in the initiatives, or in the themes.

app/(editor)/editor/page.tsx
```diff
@@ -86,6 +95,65 @@ async function getStoryData({ path }: PageProps['searchParams']): Promise<ISbRes
 
 };
 
+/**
+ * Get a list of stories that are of component sbStoryMvp in reverse chronological order.
+ */
+async function getStoryList({ path }: PageProps['searchParams']) {
```
sherakama, line 101: Adding duplicative or page-specific code to the editor is a code smell for me. I'd rather move this deeper into the component stack but before I make any suggestions can you tell me why you chose to go this way and what issues you ran into?

yvonnetangsu: @sherakama No specific reason here. I wanted to do server side fetching so just putting it where we're already fetching story data and in case we need it available for other pages in the future (e.g., initiative and theme pages might need their own story list), but definitely open to doing it in a more maintainable way 😄

app/(storyblok)/[[...slug]]/page.tsx
```diff
@@ -119,6 +128,64 @@ async function getStoryData(params: { slug: string[] }) {
   return { data: 404 };
 };
 
+/**
+ * Get a list of stories that are of component sbStoryMvp in reverse chronological order.
+ * This is used for the story filter pages.
+ */
+async function getStoryList(params: { slug: string[] }) {
```
Comment (sherakama) [inline app/(storyblok)/[[...slug]]/page.tsx:135]: You probably don't have to do the fetching here. You could move the data fetching into the storyFilterPage component, or make it a sub-component of, and do the fetching there to keep this component tidy. That could possibly also solve the editor duplicative code issue.

Comment (sherakama) [reply app/(storyblok)/[[...slug]]/page.tsx:135]: Have a look at https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#interleaving-server-and-client-components for examples and inspiration.

app/(editor)/editor/page.tsx
sherakama: Ok cool. This will be a good exercise to dig deeper into server/client component interweaving. Have a read through https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#interleaving-server-and-client-components and explore a bit more then let's reconnect to see what blockers we hit.

app/(storyblok)/[[...slug]]/page.tsx
Comment (yvonnetangsu) [reply app/(storyblok)/[[...slug]]/page.tsx:135]: @sherakama I actually did think about going that route but thought I'd do it in page since Storyblok is already initialized there 😂  If I go for the server component, do I just initialize using `StoryblokInit` again there 🤔 ?

app/(editor)/editor/page.tsx
yvonnetangsu: Thank you - I'll do that! 😄

app/(storyblok)/[[...slug]]/page.tsx
Comment (sherakama) [reply app/(storyblok)/[[...slug]]/page.tsx:135]: I don't think so, but I have run into issues with SB before. I'm pretty sure all you need is the API client 

```
```

If that has issues you can instantiate the API client yourself and get going.

Comment (yvonnetangsu) [reply app/(storyblok)/[[...slug]]/page.tsx:135]: Oh I didn't realize! I'll try that first. For some reason I thought every time I use getStoryblokApi I need to pass it the access token again so that was part of why I originally did everything in page.tsx 😂 Thanks! I'll definitely give this a go and see if I run into any issues.

Comment (yvonnetangsu) [reply app/(storyblok)/[[...slug]]/page.tsx:135]: It does looks like I need to pass it the access token again in order to fetch data in a server component. Not as straightforward as I thought 😂

yvonnetangsu:
@sherakama I think I'm going to merge this into the integration branch even though I haven't refactored the filtering. I'm only getting it in first so I can share it with design since this shows what everything actually looks like on a story filter page. As discussed, I'll work on the refactor of filtering deeper into components and won't use what I have now in production - promise 😄

PR #325, opened 2024-07-11
Always use draft for story data when in storyblok editor

yvonnetangsu:
# READY FOR REVIEW

# Summary
- For use in the (editor) route group, we want to always pull in the draft version of the story data.
- In the getStoryList function, it is ok to always use the published version (this doesn't seem to cause an API permission issue)

# Review By (Date)
- July 11

# Criticality
- 7

# Review Tasks

1. Look at code

### DIFF
--- a/utilities/data/getStoryData.ts
+++ b/utilities/data/getStoryData.ts
@@ -9,7 +9,7 @@ import { getStoryblokApi, StoryblokClient } from '@storyblok/react/rsc';
  * Make sure to not export the below functions otherwise there will be a typescript error
  * https://github.com/vercel/next.js/discussions/48724
  */
-async function getStoryData({ path, isEditor }: getStoryDataProps): Promise<ISbResult | { data: 404 }> {
+async function getStoryData({ path, isEditor = false }: getStoryDataProps): Promise<ISbResult | { data: 404 }> {
   const storyblokApi: StoryblokClient = getStoryblokApi();
   const activeEnv = process.env.NODE_ENV || 'development';
   let sbParams: ISbStoriesParams = {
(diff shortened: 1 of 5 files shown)


utilities/data/getStoryList.ts
```diff
@@ -38,7 +38,7 @@ async function getStoryList({ path }: getStoryDataProps) {
 
   // For more related documentation see app/(storyblok)/[[...slug]]/page.tsx
   const sbParams: ISbStoriesParams = {
-    version: activeEnv === 'development' ? 'draft' : 'published',
+    version: 'published',
```
yvonnetangsu, line 41: For getStoryList, I think it's ok to alway only show the published stories both in editor or in builds

utilities/data/getStoryData.ts
```diff
@@ -9,12 +9,12 @@ import { getStoryblokApi, StoryblokClient } from '@storyblok/react/rsc';
  * Make sure to not export the below functions otherwise there will be a typescript error
  * https://github.com/vercel/next.js/discussions/48724
  */
-async function getStoryData({ path }: getStoryDataProps): Promise<ISbResult | { data: 404 }> {
+async function getStoryData({ path, isEditor }: getStoryDataProps): Promise<ISbResult | { data: 404 }> {
```
sherakama, line 12: This works. My $0.02 would be to default the `isEditor` param to false so that you only have to pass in TRUE for the editor.

sherakama (approved):
Perhaps set the isEditor to default to false. This works and is fine if you dont.

utilities/data/getStoryData.ts
yvonnetangsu: Sure - will do!

yvonnetangsu:
> Perhaps set the isEditor to default to false. This works and is fine if you dont.

It's been updated - thanks @sherakama! 😸

yvonnetangsu:
Going to push this through @sherakama 🤠

PR #339, opened 2024-09-04
DS-853 | Add id options to 10 storyblok components; update CtaLink

yvonnetangsu:
# READY FOR REVIEW

# Summary
- The goal of this ticket is to allow adding ids to some components (Section was the one that was asked, but adding to more components due to the way it's used). Also, confirm that anchor link is working for CTA component and links in WYSIWYG.
- Add id option to 10 components
- If `cached_url` is home in `CtaLink`, set it to empty string (the / is added afterwards in another statement). Note: In some cases, linking to the home page will result in cached_url as `/`. Actually I believe this is the case before, but I found that sometimes it can be `home`, so adding this check to replace `home`. 

# Review By (Date)
- Retro

# Criticality
- 6

# Review Tasks

1. I've published a couple test pages to test this. They're now unpublished, but you can see the preview build here:
https://deploy-preview-339--giving-campaign.netlify.app/test/test-stories/yvonne-test-new-story-hero
Test the 3 links in the WYSIWYG with the anchor links.
Screenshot 2024-09-03 at 10 55 02 PM

2. Go to this page
https://deploy-preview-339--giving-campaign.netlify.app/test/yvonne-test-homepage-reorder-changemaker
and test the 2 link button with anchor links
Screenshot 2024-09-03 at 10 57 14 PM

- JIRA ticket(s) - DS-853

### DIFF
--- a/components/Cta/CtaLink.tsx
+++ b/components/Cta/CtaLink.tsx
@@ -47,6 +47,10 @@ export const CtaLink = React.forwardRef<HTMLAnchorElement, CtaLinkProps>(
     if (isInternal) {
       myLink = cachedUrl || href;
 
+      if (myLink === 'home') {
+        myLink = '';
+      }
+
       if (!myLink?.startsWith('/')) {
         myLink = `/${myLink}`;
       }
--- a/components/Storyblok/SbBasicCard.tsx
+++ b/components/Storyblok/SbBasicCard.tsx
@@ -14,6 +14,7 @@ import { type SbImageType } from './Storyblok.types';
 export type SbBasicCardProps = {
   blok: {
     _uid: string;
+    id: string;
     superhead?: string;
     heading?: string;
     headingLevel?: HeadingType;
@@ -35,6 +36,7 @@ export type SbBasicCardProps = {
 
 export const SbBasicCard = ({
   blok: {
+    id,
     superhead,
     heading,
     headingLevel,
@@ -66,6 +68,7 @@ export const SbBasicCard = ({
   return (
     <BasicCard
       {...storyblokEditable(blok)}
+      id={id}
       superhead={superhead}
       heading={heading}
       headingLevel={headingLevel || 'h3'}
--- a/components/GAProvider.tsx
+++ b/components/GAProvider.tsx
@@ -1,5 +1,5 @@
 'use client';
-import { GoogleTagManager } from '@next/third-parties/google'
+import { GoogleTagManager } from '@next/third-parties/google';
 import { useEffect } from 'react';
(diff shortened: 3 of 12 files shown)


components/Cta/CtaLink.tsx
```diff
@@ -47,6 +47,10 @@ export const CtaLink = React.forwardRef<HTMLAnchorElement, CtaLinkProps>(
     if (isInternal) {
       myLink = cachedUrl || href;
 
+      if (myLink === 'home') {
+        myLink = '';
```
sherakama, line 51: Shouldn't this be `/` instead of an empty string?

yvonnetangsu: It could be either since the conditional right below it adds the `/ `if it doesn't start with one already 😄 
I didn't expect a review this fast 😂

components/Storyblok/SbBasicCard.tsx
```diff
@@ -14,6 +14,7 @@ import { type SbImageType } from './Storyblok.types';
 export type SbBasicCardProps = {
   blok: {
     _uid: string;
+    id: string;
```
sherakama, line 17: I vaguely remember something about the `id` in storyblok's response not being a good fit for an HTML element ID. I am trying to remember if that was because it was not unique or if it was because the contents were not a valid id name...

Perhaps it was both. 

I'd have to look at the API/docs to find out but perhaps you swap this out for a `useId()` call instead?

yvonnetangsu: @sherakama Hmmm I don't seem to remember the id field causing problems. We still have it for our other Storyblok sites 👀  I did add a regex check to not let the content author use the id "main-content" (that was I think Alumni) so I can add that check in Storyblok for this site too. 

useId doesn't work well here because the content author will add the id manually when they want to link to a section (from a newsletter for example).

yvonnetangsu: Oh I see what you were thinking @sherakama - these are actually new id fields that I manually added to the Storyblok components for the content authors to input themselves, so not the ids that are automatically associated with each components in the Storyblok response 🫡

sherakama: Oh, you created `id` fields? Sounds risky but you have power users so... ok.

yvonnetangsu: Yes. I added instructions and regex check 😆

yvonnetangsu:
Thank you for your eyes @sherakama 🤠 I'm going to push this live since they need it before Sep 17 for the newsletter.

PR #341, opened 2024-09-05
NoJira: Add embed label if there is one to the typeform iframe.

sherakama:
# READY FOR REVIEW

# Summary
- Adds an option for content authors to enter an aria-label for the iframe

# Review By (Date)
- Whenever

# Criticality
- Normal

# Review Tasks

1. Review https://momentum.stanford.edu/newsletter/sign-up and inspect typeform iframe
2. Check out this branch on your local
3. Build and review the newsletter sign up page
4. Review code and provide comments
5. Navigate to https://app.storyblok.com/#/me/spaces/1005200/stories/0/0/545632/blok/61d7aec6-9820-4e8b-a424-d074b334491d
6. Click into the typeform component to show the fields in the sidebar
7. Review new field label and description words for clarity and anything else you'd like to comment on


components/Storyblok/SbTypeform.tsx
```diff
@@ -62,6 +64,9 @@ export const SbTypeform = ({
   const height = parseInt(embedHeight, 10);
   const resize = autoResize || !height;
   const opacity = transparent ? 0 : 100;
+  const iframeProps = {
+    ariaLabel: embedLabel ?? undefined,
```
yvonnetangsu, line 68: @sherakama Thank you so much for taking care of this request! Only tiny tweak I would ask for is this line
```suggestion
    ariaLabel: embedLabel ? embedLabel : undefined,
```
The `??` doesn't count empty string as false. Since `embedLabel` is default to `''`, if the field in Storyblok is empty, it will still add `aria-label=""` (or sometimes appear just as `aria-label`) 😄 which I don't think is what you want.

Screenshot 2024-09-05 at 1 33 42 PM

yvonnetangsu (changes requested):
Just one tiny request otherwise everything looks great 👍🏼 🚀 Storyblok field looks good to me too!

components/Storyblok/SbTypeform.tsx
sherakama: Oh shoot yeah. The NULLcoalesce operator should have been `||` as that does truthy.

yvonnetangsu: That got me all the time too 😆

yvonnetangsu (approved):
Thank you for supersonic speed PR 🚀 😄

PR #362, opened 2024-10-17
Nojira | More robust URL processing; prevent empty redirect entries causing build error

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Use `sbStripSlugUrl` utility to remove prefix from internal links, but add more robust href processing

# Review By (Date)
- Soon

# Criticality
- 5

# Review Tasks

1. Go to the preview
2. Click some links in the masthead
3. Click some story cards
4. Click on a taxonomy chip to go to a story list page
5. Click on some story cards there
6. Go to the 404 page and check that CTA goes back to the homepage
7. Clicky clicky some more
8. Check that the sitemap still works the same https://deploy-preview-362--giving-campaign.netlify.app/sitemap.xml
9. Inspect the metadata on a couple pages and check that canonical URL etc still works properly
10. Look at code


components/Storyblok/SbStoryCard.tsx
```diff
@@ -85,7 +85,7 @@ export const SbStoryCard = ({
       imageFocus={cardFocus || storyCardFocus || heroFocus || bgFocus}
       tabColor={paletteAccentColors[value || storyTabColorValue]}
       link={link}
-      href={`/${full_slug}`}
+      href={full_slug}
```
yvonnetangsu, line 88: We remove all leading slash before we check for the prefix in the updated sbStripSlugUrl utility so we don't need to add a leading slash here anymore

utilities/sbStripSlugUrl.ts
```diff
@@ -31,7 +31,25 @@ export const sbStripSlugURL = (url:string): string => {
...
   }
 
-  const parts = url.split('/');
+  // Remove the leading slash from the URL
+  const urlNoLeadingSlash = url.replace(/^\//, '');
+
+  /**
+   * If the URL is just the prefix, return a slash.
+   * Occasionally the cached_url for the momentum homepage still shows up as
+   * 'home' so adding the last check just in case.
```
sherakama, line 40: Where is `home` coming from? The `real_path`?

components/Cta/CtaNextLink.tsx
```diff
@@ -32,27 +33,14 @@ export const CtaNextLink = React.forwardRef<HTMLAnchorElement, CtaNextLinkProps>
...
-  const cleanParts = hrefParts.filter((s:string) => s.length);
-
-  // If the first part of the URL is the slug prefix, remove it.
-  if (cleanParts[0] === prefix) {
-    cleanParts.shift();
-  }
-
-  const strippedHref = `/${cleanParts.join('/')}`;
+  const strippedHref = sbStripSlugURL(href);
+  const processedHref = strippedHref.startsWith('/') ? strippedHref : `/${strippedHref}`;
```
sherakama, line 37: I think we should change `sbStripSlugURL` to always return with the leading slash instead of trying to handle this condition outside of it everywhere else.

yvonnetangsu: Sure I can do that!

utilities/sbStripSlugUrl.ts
yvonnetangsu: I think it's somewhere in the cache 😬 It only happens sometimes in Storyblok (they have some weird caching issue when on their site occasionally). When I looked at the draft JSON, "home" was not the slug, but I've seen it rendered a few times in the editor so just putting it here just in case.

sherakama: Huh. Ok. Thanks for explaining.

yvonnetangsu:
@sherakama I just found something weird, but maybe you know this already? I guess the `cached_url` really is cached, but for some reason I remember it changing when a story is moved in the past 🤔 

I did an experiment and went to a test page I added a long time ago before we did the directory split. There's an existing CTA to the changemaker page (on the left). I added a new one on the right that links to the same changemaker page. See how they have different `cached_url` - the old one is still changemaker but the new one has `cached_url` momentum/changemaker 👀  I've already opened up the old CTA on the left to jiggle it to make sure it has updated itself before I saved and looked at the draft JSON.

Screenshot 2024-10-17 at 10 48 40 AM
Screenshot 2024-10-17 at 10 48 33 AM

utilities/data/getStoryblokRedirects.mjs
```diff
@@ -100,12 +100,13 @@ export const getStoryblokRedirects = async () => {
...
 
-  const redirects = stories.map(entry => ({
+  const redirects = stories
+  .filter(entry => entry.content.from && entry.content.to) // Filter out entries without "from" or "to"
+  .map(entry => ({
     source: sanitizeSourcePath(entry.content.from),
     has: extractQueryParameters(entry.content.from),
     destination: entry.content.to,
-    statusCode: sanitizeRedirectCode(entry.content.statusCode),
+    statusCode: sanitizeRedirectCode(entry.content.statusCode || 301),
```
yvonnetangsu, line 109: @sherakama Something interesting about creating redirects - Even though the `from` and `to` fields in Storyblok are required, and technically you can't save a redirect without entering something in those fields, the entry still exists without any values in those fields the moment it's created, so the empty test entry that you created in the demo today actually got "saved" and held up the Netlify build by causing an error 😆 Added this to filter out any entries without `to` and `from` fields, and add a fall back status code 301 here.

Screenshot 2024-10-17 at 2 01 33 PM

components/Storyblok/SbStoryListNav/SbStoryListNav.tsx
```diff
@@ -26,21 +26,23 @@ type SbStoryListNavType = {
...
   return (
     <ul className={styles.list}>
-      {links.map((link) => {
-        const cached_url = link.link?.cached_url ? sbStripSlugURL(link.link.cached_url) : '';
-        const isCurrentPage = fullSlug === cached_url || (fullSlug === 'stories' && cached_url === 'stories/');
+      {links.map(({ _uid, label, link }) => {
+        const processedCachedUrl = sbStripSlugURL(link?.cached_url || '');
+        const processedFullSlug = sbStripSlugURL(fullSlug);
+        const isCurrentPage = processedFullSlug === processedCachedUrl;
+
```
yvonnetangsu, line 33: After I updated the sbStripSlugURL function, the current page highlight in the story filter page nav stop working, so I made some more tweaks to the function, and process both the full slug and the cached_url so that they're 100% in the same format before comparing them

yvonnetangsu:
@sherakama This should be working as expected now. Please retest the following especially:

1. Click some links on the page
2. Click some taxonomy chips
3. While on the story filter page, check that the current page nav pill in the nav is highlighted red
4. Click the "All" story filter menu nav item and check that it gets highlighted once it lands on the page
5. Check the sitemap.xml again
6. Check the canonical URLs
7. Look at the code updates

utilities/data/getStoryblokRedirects.mjs
sherakama: Oh that is kinda ridiculous. More Storyblok fun!

sherakama:
> @sherakama I just found something weird, but maybe you know this already? I guess the `cached_url` really is cached, but for some reason I remember it changing when a story is moved in the past 🤔
> 
> I did an experiment and went to a test page I added a long time ago before we did the directory split. There's an existing CTA to the changemaker page (on the left). I added a new one on the right that links to the same changemaker page. See how they have different `cached_url` - the old one is still changemaker but the new one has `cached_url` momentum/changemaker 👀 I've already opened up the old CTA on the left to jiggle it to make sure it has updated itself before I saved and looked at the draft JSON.
> 
> Screenshot 2024-10-17 at 10 48 40 AM Screenshot 2024-10-17 at 10 48 33 AM

Huh, yeah. 

This is in alignment of Storyblok's 'Dont touch the content' when config changes approach. I would have expected the URL to have changed to the referenced story however. 

This case was supposed to have been mitigated by the URL strip function and the checking for the slug first before lopping off the first part. 

Oh man.

utilities/data/getStoryblokRedirects.mjs
yvonnetangsu: I added an empty redirect to test and this PR preview built alright after I pushed up the commit to filter out those entries. I have now deleted that draft empty entry because it blocks my other PR preview 😂

components/Storyblok/SbStoryListNav/SbStoryListNav.tsx
sherakama: Aye, the forward and leading slashes get handled this way. 👍

utilities/data/getStoryblokRedirects.mjs
```diff
@@ -49,7 +49,7 @@ const sanitizeRedirectCode = (code) => {
  * @returns {string}
  */
 const sanitizeSourcePath = (str) => {
-  const parts = str.split('?');
+  const parts = str?.split('?');
```
sherakama, line 52: If `str` isn't a string, it shouldn't have been passed in. I feel that the optional chaining operator should not be here.

yvonnetangsu: Yeah that was while I was troubleshooting while I first got the built error because of the empty redirects. The superficial error points to the split so this was my first attempt. I think we don't need these now - I'll remove.

utilities/data/getStoryblokRedirects.mjs
sherakama, line 109: You shouldn't need the `|| 301` here as the sanitize function should default to that already
https://github.com/SU-SWS/ood-giving-campaign/blob/003d5de75eae1ad8d58438e373728b65489d7dc5/utilities/data/getStoryblokRedirects.mjs#L43

sherakama (approved):
A couple of pendantic code style comments and then you're GTG. Thanks for this work.

yvonnetangsu:
Thank you for the review @sherakama. I'm going to merge my annotated image PR into `dev` as well (just style and responsive image updates and Alexis has reviewed). I think I have a 3rd PR for some Momentum updates - after that I'll cut a release 🤠

PR #389, opened 2025-02-06
DS-1146 DS-1193 | Looping video page heroes and inline Story Video component; reusable Caption, MediaWrapper and various Video components

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Add background and foreground video looping to story page heroes
- Add `SbVideo` - inline looping video component with caption and other options
- Add `Video`, `MutedVideoLoop`, `StoryVideo` and `VideoButton` components
- Create `MediaWrapper` and `Caption` components which replace part of the common code in `StoryImage`, `StoryVideo` and `EmbedMedia` components

# Review By (Date)
- Sooner the better

# Criticality
- 6

# Review Tasks

1. Review pages in Slack conversation:
https://stanfordwebservices.slack.com/archives/C03R765SAJC/p1739990416255369
https://stanfordwebservices.slack.com/archives/C03R765SAJC/p1740163729849879
2. Look at the deploy preview and make sure that the story image and embed media components still work the same as before:
https://deploy-preview-389--giving-campaign.netlify.app/initiatives/stanford-public-humanities
https://deploy-preview-389--giving-campaign.netlify.app/stories/team-behind-the-teams (The Annotated Image component uses StoryImage)
https://deploy-preview-389--giving-campaign.netlify.app/stories/stretching-the-canvas
3. Look at code and things called out

- JIRA ticket(s) - DS-1146, DS-1193


utilities/getProcessedImage.ts
```diff
@@ -16,7 +19,7 @@ import { getMaskedAsset } from './getMaskedAsset';
 
 export const getProcessedImage = (
   imageSrc: string = '',
-  crop: string = '',
+  crop: string = getSbImageSize(imageSrc)?.width > 2000 ? '2000x0' : '',
```
yvonnetangsu, line 22: Made a small change to getProcessedImage utility, so that If a crop dimension is not provided and if the image width is greater than 2000px, the crop dimension will be set to "2000x0" to keep the original aspect ratio and downsize the image to 2000px wide. I don't think there is a reason to have any images on the site that is larger than 2000px wide at the moment.

components/Video/Video.tsx
```diff
@@ -0,0 +1,33 @@
...
+    ref={ref}
+    poster={getProcessedImage(posterSrc)}
+    className={className}
+    {...props}
+  >
+    {webmSrc && <source src={getMaskedAsset(webmSrc)} type="video/webm" />}
+    {mp4Src && <source src={getMaskedAsset(mp4Src)} type="video/mp4" />}
+    {children}
+  </video>
+));
```
yvonnetangsu, line 33: Generic html video component that supports a mp4 and webm sources (plus any additional via `children` prop), and a preview image `poster` attribute

components/Video/MutedVideoLoop.tsx
```diff
@@ -0,0 +1,10 @@
+import { forwardRef } from 'react';
+import { Video, type VideoProps } from './Video';
+
+/**
+ * This component is a more specific version of the Video component created for convenience.
+ * It is used for decorative muted video loops.
+ */
+export const MutedVideoLoop = forwardRef<HTMLVideoElement, VideoProps>((props, ref) => (
+  <Video ref={ref} role="presentation" muted autoPlay loop playsInline {...props} />
+));
```
yvonnetangsu, line 10: A convenience component for muted loops so I don't have to write out those same attributes all the time.

components/Video/StoryVideo.tsx
```diff
@@ -0,0 +1,125 @@
...
+  videoPosterSrc?: string;
+  isFullScreen?: boolean; // Whether the video takes the full width of the device
+  spacingTop?: PaddingType;
+  spacingBottom?: PaddingType;
+  animation?: AnimationType;
+  delay?: number;
+  aspectRatioClass?: string;
+};
+
+export const StoryVideo = ({
```
yvonnetangsu, line 28: This is a counterpart of the `StoryImage` component - comes with a wrapper that provides aspect ratio, width and spacing options, plus caption and other options.

components/Media/Caption.tsx
```diff
@@ -0,0 +1,36 @@
...
+  // Inset the caption to centered container width when the media is full screen width
+  isCaptionInset?: boolean;
+  captionBgColor?: styles.CaptionBgColorType;
+};
+
+export const Caption = ({
+  caption,
+  isCaptionInset,
+  captionBgColor = 'transparent',
+  className,
```
yvonnetangsu, line 21: Extract the Caption component since it's used in a few places

components/Media/MediaWrapper.tsx
```diff
@@ -0,0 +1,70 @@
...
+  boundingWidth?: 'site' | 'full';
+  width?: WidthType;
+  pt?: PaddingType;
+  pb?: PaddingType;
+  animation?: AnimationType;
+  delay?: number;
+  aspectRatioClass?: string; // Additional eg, responsive aspect ratio classes
+};
+
+export const MediaWrapper = ({
```
yvonnetangsu, line 27: This is the wrapper component that's used for StoryImage and StoryVideo that provides it with a common set of options

components/Video/StoryVideo.tsx
```diff
@@ -0,0 +1,125 @@
...
+  useEffect(() => {
+    const video = videoRef.current;
+    if (!video) return;
+
+    if (isVideoInView && !isUserPaused) {
+      video.play().catch(() => {});
+    } else {
+      video.pause();
+    }
+  }, [isVideoInView, isUserPaused]);
```
yvonnetangsu, line 84: Pause the video when 90% or more is out of the viewport. Resumes playing when it comes back into view, unless user had previously manually paused the video.

components/Animate/AnimateInView.tsx
```diff
@@ -45,7 +45,7 @@ export const AnimateInView = ({
         duration,
         ease: 'easeOut',
       }}
-      initial="hidden"
+      initial={beforeAnimationState}
```
yvonnetangsu, line 48: Updated AnimateInView wrapper. This still works as intended on live, but when this component is used with a newer version of Framer Motion, it doesn't work correctly. This get rid of the conflict and should be this way from the beginning - I thought this looked sus at one point and always meant to come back to do it properly even though it was still working.

components/BlurryPoster/BlurryPoster.tsx
```diff
@@ -95,6 +109,53 @@ export const BlurryPoster = ({
...
+  const [isBgPlaying, setIsBgPlaying] = useState<boolean>(false);
+  const [isBgUserPaused, setIsBgUserPaused] = useState<boolean>(false);
+
+  // Toggle foreground video play/pause
+  const toggleBgVideo = () => {
+    if (!bgVideoRef.current) return;
+
+    setIsBgPlaying((prev) => {
+      if (prev) {
+        bgVideoRef.current?.pause();
```
sherakama, line 133: do you need the `?` here as you've already checked if the `.current` value is truthy?

components/BlurryPoster/BlurryPoster.tsx
```diff
@@ -95,6 +109,53 @@ export const BlurryPoster = ({
...
+        setIsBgUserPaused(true);
+      } else {
+        bgVideoRef.current
+          ?.play()
+          .catch(() => {});
```
sherakama, line 138: Tell me more about this catch.

components/BlurryPoster/BlurryPoster.tsx
yvonnetangsu: Good catch! I'll update 😄

utilities/getProcessedImage.ts
sherakama: Smart.

components/Animate/AnimateInView.tsx
yvonnetangsu: The reason I found the bug was that Rebecca is using my AnimateInView component on Centennial which is on React 19/Next 15 and the latest version of Framer Motion (now called just Motion). She was wondering why the reduced motion setting didn't work for her site, but still works for Momentum. I came back to look at this and found that it's fixed if I make this change. @sherakama

components/BlurryPoster/BlurryPoster.tsx
yvonnetangsu: I added this because sometimes video playback might be blocked or it can't be played for some reason, which can cause issues with DOM errors. This catches the error silently and prevents that from causing issue - I could have it display a console error or warning as well.

yvonnetangsu: Honestly I can certainly do more with error handling, but error should be rare. So in the interest of time, I'll come back to this later @sherakama 😂

sherakama:
Happy path testing works (and looks) great. 
You've done some good work on abstracting and consolidating. Looks like there is potential for some more as there is a little code duplication going on here between the image and video components but nothing that should be a blocker IMHO. 

I added @ericbakenhus for visibility and his keen eye too.

components/BlurryPoster/BlurryPoster.tsx
sherakama: I'll let you judge the time/effort to spend here but overall I am not a fan of squashing errors quietly as they tend to be very difficult to troubleshoot. I'd rather see the error logged.

yvonnetangsu:
> Happy path testing works (and looks) great. You've done some good work on abstracting and consolidating. Looks like there is potential for some more as there is a little code duplication going on here between the image and video components but nothing that should be a blocker IMHO.
> 
> I added @ericbakenhus for visibility and his keen eye too.

Yes, I remember what you suggested before about a responsive image component 😄  I still plan on doing that at some point, but going to get this out the door first. It's a busy sprint 🤠 Some finally clean up of redundant props coming (get rid of those that are already in MediaWrapperProps)

yvonnetangsu:
I'll have to find out more about this "happy path testing" @sherakama 👀

components/BlurryPoster/BlurryPoster.tsx
yvonnetangsu: I can update for it to show a console warning 👍🏼

sherakama:
> I'll have to find out more about this "happy path testing" @sherakama 👀

Well, not a long path on this one, but the testing was:
1. Load page
2. See video play
3. Resize page
4. Pause 
5. Reize page
6. Play

yvonnetangsu:
> > I'll have to find out more about this "happy path testing" @sherakama 👀
> 
> Well, not a long path on this one, but the testing was:
> 
> 1. Load page
> 2. See video play
> 3. Resize page
> 4. Pause
> 5. Reize page
> 6. Play

Cool 😎 Yeah I've never heard that term before so I just looked that up.

PR #397, opened 2025-03-11
DS-1147 | Quote alignment updates; metadata update to add dek as fallback for description

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Teaser text starts on a new line instead of be on the same line as the quotation mark
- Spacing finetune
- Fix bug where metadata isn't using story dek as fallback description

# Review By (Date)
- Soon

# Criticality
- 6

# Review Tasks

## Setup tasks and/or behavior to test

1. Go to https://deploy-preview-397--giving-campaign.netlify.app/test/yvonne-test-quote-align
2. Check that in the variant where there is a large quote mark at the top, it's on its own line instead of being on the same line as the teaser.
3. Go to https://deploy-preview-397--giving-campaign.netlify.app/stories/education-for-a-life-of-purpose
4. Check the HTML head, that the description, og:description and twitter:description all says "Civil society is fraying a bit. COLLEGE has a fix."
Screenshot 2025-03-11 at 4 27 58 PM

# Associated Issues and/or People
- JIRA ticket(s) - DS-1147

PR #396, opened 2025-03-11
DS-1148 | Support 2-up Changemaker Cards

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Use container query so that Changemaker Cards look good in a 2-col grid for all breakpoints
- Modify responsive aspect ratio of horizontal changemaker cards in the mid-range breakpoints

# Review By (Date)
- Retro

# Criticality
- 5

# Review Tasks

## Setup tasks and/or behavior to test

1. Go to https://deploy-preview-396--giving-campaign.netlify.app/test/test-changemakers
2. Look at the 2-up horizontal, 1-up horizontal and vertical changemaker cards and check the font size and everything looks proportional at all breakpoints

# Associated Issues and/or People
- JIRA ticket(s) - DS-1148

PR #398, opened 2025-03-11
Release/v3.7.2

yvonnetangsu:
# READY FOR REVIEW

# Summary
- #396 
- #397 

# Review By (Date)
- Retro

# Criticality
- 6

# Review Tasks

1. See #396 amd #397


components/Storyblok/SbQuote.tsx
```diff
@@ -58,7 +58,7 @@ export const SbQuote = ({
       isLargeBody={isLargeBody}
       source={source}
       barColor={paletteAccentColors[value]}
-      quoteOnRight={quoteOnRight}
+      barOnLeft={quoteOnRight}
```
sherakama, line 61: Why not change the Storyblok field name too?

components/Quote/Quote.tsx
```diff
@@ -76,24 +77,24 @@ export const Quote = ({
...
+          {body && (
+            <Text
+              weight="semibold"
+              variant={isLargeBody ? undefined : 'big'}
+              leading="display"
+              size={isLargeBody ? 2 : undefined}
+              font="serif"
+              className={styles.body(hasRealTeaser)}
+            >
+              {body}
```
sherakama, line 90: Is body not a rich text field anymore? How'd you handle the transition?

components/Storyblok/SbQuote.tsx
yvonnetangsu: Because I want to keep the backward compatibility. The quote is used in many places so I'd rather not have to manually go in and change them all out in Storyblok if I rename the field.

components/Quote/Quote.tsx
yvonnetangsu: This one was never a rich text field. I think during the POC stage, it was undecided, and we never removed the unnecessary support for rich text later on.

components/Storyblok/SbQuote.tsx
sherakama: Eh? Wouldn't you just have to change the one Storyblok blok (Quote)? Changing the field name keeps the data as far as I recall.

yvonnetangsu: It does keeps the data until a user opens up a story with a quote on it. Then it will empty out the selection that was picked due to the field name change. I noticed that on this site, they do go back and update old stories a lot, and they might not know to check existing quotes whether the option they have selected has changed 😬

sherakama: Wait what? That seems like a very bad bug in Storyblok's UI.

yvonnetangsu: We almost never update a field name after a component has been deployed to live, so I'm relying on memory of how it once was. Next time I can test this out before we deploy a new component 😂 For now, I'd rather not risk this one.

PR #402, opened 2025-03-17
DS-856 | Accordion

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Accordion component

# Review By (Date)
- Retro

# Criticality
- 4

# Review Tasks

1. Go to https://deploy-preview-402--giving-campaign.netlify.app/test/yvonne-test-accordion
2. Look at both the light and dark mode accordion
3. Test all the functionality including using keyboard. You should be able to tab from one item to the other and open/close with spacebar/enter key
4. Check responsive behavior
5. Check that it follows the a11y pattern recommended here:
https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/
6. Look at code

- JIRA ticket(s) - DS-856


sherakama:
@yvonnetangsu 

This is looking good. There is a small bug with the keyboard navigation where it focuses on the hidden content on the first expand/collapse all

Steps to reproduce
1. Load the page
2. Tab navigate down to the first expand all button
3. Hit return to expand all
4. tab nav into the first collapsible heading
5. Hit return to close the first accordion
6. Hit tab
7. Focus goes to link in hidden content instead of the next accordion heading

components/Cta/Cta.styles.ts
```diff
@@ -38,8 +38,8 @@ export const ctaVariants = {
   close: 'inline-block font-semibold leading-none text-digital-red-light hocus:text-digital-red-xlight focus:outline-none',
   'close-x': 'leading-none',
   chip: 'inline-block leading-display no-underline text-current rounded-full border-2 border-current hocus:text-current font-normal underline-offset-4 decoration-transparent hocus-visible:decoration-current hocus-visible:decoration-2',
-  storyCardChip: `${chipBase} ${chipLight}`,
-  storyCardChipDark: `${chipBase} ${chipDark}`,
+  storyCardChip: `story-tag ${chipBase} ${chipLight}`,
+  storyCardChipDark: `story-tag ${chipBase} ${chipDark}`,
```
yvonnetangsu, line 42: Added the story-tag class for Forrest for analytics here @sherakama

components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,160 @@
...
+              }}
+              initial={false}
+              transition={{
+                height: { duration: 0.2, ease: 'easeIn' },
+                /**
+                 * Use a shorter duration for visibility to prevent links in content being focusable
+                 * while the content is animating out while collapsing
+                 */
+                visibility: { duration: 0.1 },
+              }}
```
yvonnetangsu, line 140: @sherakama I suspected that the issue you experienced was due to the original 0.3 sec duration when the visibility animates from hidden to visible (visibility hidden is what makes the content un-focusable). I have lower the duration to 0.1 sec. Previously I occasionally was able to reproduce what you experienced, but now it seems to be ok for me. Please retest and let me know 😄

components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,154 @@
...
+              animate={{
+                height: openItems[index] ? 'auto' : 0,
+                visibility: openItems[index] ? 'visible' : 'hidden',
+              }}
+              initial={false}
+              transition={{ duration: 0.3, ease: 'easeIn' }}
+              className={styles.contentWrapper}
+            >
+              {/* @ts-expect-error Reason: inert not fully support in React 18. Using '' sets it to true */}
+              <div className={styles.richtextWrapper} {...(!openItems[index] ? { inert: '' } : {})}>
```
yvonnetangsu, line 137: @sherakama I found a better way adding the inert attribute to make the child items un-focusable. However, React 18 doesn't fully support this attribute, so if you set `inert={true}` (or just `inert`) it doesn't get rendered in the DOM. A workaround is to set `inert=''` instead of just adding `inert` as a boolean true attribute.
https://github.com/facebook/react/issues/17157

components/Accordion/Accordion.tsx
yvonnetangsu: Found a better way to deal with it
https://github.com/SU-SWS/ood-giving-campaign/pull/402#discussion_r2004401640

components/Accordion/Accordion.tsx
sherakama: Neat. Thanks for sharing that learning.

sherakama (approved):
GTG. Thank you!

yvonnetangsu:
Thanks for the review @sherakama !

PR #407, opened 2025-03-25
Nojira | Sync tour changes

yvonnetangsu:
# READY FOR REVIEW

# Summary
- While I was sync-ing Momentum changes to Tour, I discovered small things that I want to bring from Tour back to Momentum.
- Update `getProcessedImage` to not use Storyblok image service stuff on svgs
- Update Event Banner (a11y and style fix up)
- Update `@types/node`

# Review By (Date)
- Retro

# Criticality
- 4

# Review Tasks

1. Look at code and preview
2. Go to https://deploy-preview-407--giving-campaign.netlify.app/experience-the-moment
3. See that the druk headings are left aligned whether there is a date or not on the event banner
Screenshot 2025-03-25 at 2 55 37 PM
4. Inspect the superheading and see that it's a h2 with SR only text of the Druk heading following the visible superheading text.

### DIFF
--- a/utilities/getProcessedImage.ts
+++ b/utilities/getProcessedImage.ts
@@ -25,6 +25,9 @@ export const getProcessedImage = (
 ): string => {
   if (!imageSrc) {
     return '';
+  // Leave the SVGs alone since they don't need to be processed
+  } else if (imageSrc.endsWith('.svg')) {
+    return getMaskedAsset(imageSrc);
   }
 
   // Get the width and the height from the crop dimension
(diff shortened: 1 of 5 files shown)


utilities/getProcessedImage.ts
```diff
@@ -25,6 +25,9 @@ export const getProcessedImage = (
 ): string => {
   if (!imageSrc) {
     return '';
+  // Leave the SVGs alone since they don't need to be processed
+  } else if (imageSrc.endsWith('.svg')) {
+    return getMaskedAsset(imageSrc);
```
yvonnetangsu, line 30: This was added on Tour because their logo is svg. It would be useful on Momentum too.

yvonnetangsu: @ericbakenhus Updating `getProcessedImage` utility here 😄

PR #411, opened 2025-04-29
A11y | DS-1510 | Add role="button" to clickable video preview wrapper

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Add `role="button"` to the clickable div when EmbedMedia preview mode is on (I can't change the HTML element so this is a compromise to make it accessible)
- Better handling when space key is pressed while focused on the preview (so it doesn't scroll down the page - should behave like enter key)
- Update `getNumBloks`

# Review By (Date)
- Soon

# Criticality
- 5

# Review Tasks

1. Go to https://deploy-preview-411--giving-campaign.netlify.app/test/yvonne-test-video-preview-a11y
2. There are 3 embed media on the page. Only the first 2 are using the preview mode. The 2nd one has a custom preview image while the 1st one uses the default. Inspect the first two players using the preview mode. Check that the `div` has `role="button"`  as an attribute.
Screenshot 2025-04-29 at 4 09 14 PM

3. Press the space bar while you're focused on the preview, check that it plays the video and doesn't scroll the page.
4. Inspect the last embed player on the page. It shouldn't have any role="button" class because it's not using the preview mode.
5. Look at code

- JIRA ticket(s) - DS-1510

### DIFF
--- a/components/EmbedMedia/EmbedMedia.tsx
+++ b/components/EmbedMedia/EmbedMedia.tsx
@@ -1,6 +1,7 @@
 import React, { useEffect, useRef, useState } from 'react';
 import { cnb } from 'cnbuilder';
 import ReactPlayer from 'react-player/lazy';
+import { useEventListener } from 'usehooks-ts';
 import { Caption, type CaptionProps } from '@/components/Media/Caption';
 import { FlexBox } from '@/components/FlexBox';
 import { PreviewImage } from './PreviewImage';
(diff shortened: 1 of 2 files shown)


components/EmbedMedia/EmbedMedia.tsx
```diff
@@ -49,13 +50,48 @@ export const EmbedMedia = ({
...
-      setHasWindow(true);
-    }
+    setHasWindow(typeof window !== 'undefined');
   }, []);
 
-  const PreviewImg = previewImageSrc ? <PreviewImage previewImageSrc={previewImageSrc} /> : null;
+  // If the space key is pressed, do not scroll the page and click the element
+  const handleSpaceKeyDown = (event: KeyboardEvent) => {
+    if (event.code === 'Space' || event.key === ' ') {
+      event.preventDefault();
```
yvonnetangsu, line 63: This prevents scrolling down the page when the spacebar is clicked

components/EmbedMedia/EmbedMedia.tsx
```diff
@@ -49,13 +50,48 @@ export const EmbedMedia = ({
...
+
+    const wrapper = playerWrapperRef.current;
+
+    /**
+     * Observe for an element with class .react-player__preview inside the wrapper
+     * and set its role to button when it is added to the DOM
+     */
+    const observer = new MutationObserver(() => {
+      const preview = wrapper.querySelector('.react-player__preview');
+      if (preview) {
```
sherakama, line 79: Do you need to check `if(preview)`? Does the element show up otherwise? Even if it does, do you care?

sherakama (approved):
Looks and functions well. 

It's not every day you get to use the MutationObserver. Fun workaround. 

Screenshot 2025-04-29 at 17 07 02

yvonnetangsu:
> Looks and functions well.
> 
> It's not every day you get to use the MutationObserver. Fun workaround.
> 

Thanks for testing @sherakama ! Yeah I've never used it before. I tried some other approaches but this is the only one that works reliably 😄

yvonnetangsu:
Asking Caryl to test in JIRA so will leave this up in case she has anything to add.

yvonnetangsu:
Caryl said no more siteimprove issues. Merging in and going to do the same on Open Minds 😄

PR #421, opened 2025-07-09
NoJira: Localhost https moar better.

sherakama:
# READY FOR REVIEW

# Summary
- Use Nextjs https localhost

# Review By (Date)
- Whenever

# Criticality
- none

# Review Tasks

Follow instructions in docs


.nvmrc
```diff
@@ -1 +1 @@
-20
\ No newline at end of file
+22
```
sherakama, line 1: 20 is EOL. We should be on 22 now.

package.json
```diff
@@ -19,12 +19,10 @@
   ],
   "scripts": {
     "dev": "next dev",
+    "dev:https": "next dev --experimental-https --port 3010",
```
sherakama, line 22: Exists in nextjs 14

sherakama:
I seem to have broken some things with the node update. I can come back and look at that or we can revert back to 20.

yvonnetangsu:
> I seem to have broken some things with the node update. I can come back and look at that or we can revert back to 20.

I don't have any Momentum ticket for now so it's up to you (I pushed the release right before you open this PR 😆 )

package.json
```diff
@@ -67,12 +65,10 @@
     "netlify-cli": "^17.37.1",
     "netlify-plugin-vault-variables": "^3.0.0",
     "postcss": "^8.5.3",
+    "tailwindcss": "^3.4.17",
```
sherakama, line 68: This was sneaking up to 4 so I locked it here to 3

tsconfig.json
```diff
@@ -1,20 +1,15 @@
 {
   "compilerOptions": {
-    "target": "es2021",
-    "lib": [
-      "dom",
-      "dom.iterable",
-      "esnext"
-    ],
+    "target": "ESNext",
```
yvonnetangsu, line 3: I wonder if it's better to target a specific version rather than ESNext here. Maybe ES2023 if we're updating to node 22?
https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping

sherakama: Also interestingly from your link 

> After TypeScript 5.9 is released, it is recommended to set it to node20 instead of `ESXXXX`

sherakama:
@yvonnetangsu I changed the target to ES2020. 

Let me know if you'd like to see any other changes.

yvonnetangsu:
I'll merge this into dev so I can use this to work on my new ticket 🤠

